PR #13, opened 2015-07-30
Remove file_exists, is not needed

jesperveldhuizen:
Very minor but: realpath does the checking already and returns false when the file does not exists or an other error occurs.


lvgunst:
Good catch. Thank you!

PR #15, opened 2015-11-06
Add support for OAuth access

RickWong:
Add $client->setAccessToken() for OAuth Access Token support.
Add several new OAuth-only resources, objects, and examples.
Add Undefined resource for continuous forward-compatibility.


README.mdown
```diff
@@ -95,7 +95,7 @@ Retrieve the list of issuers:
 ```
 
 _`$issuers` will be a list of `Mollie_API_Object_Issuer` objects. Use the property `$id` of this object in the
- API call, and the property `$name` for displaying the issuer to your customer. For a more in-depth example, see [Example 4](https://github.com/mollie/mollie-api-php/blob/master/examples/4-ideal-payment.php)._
+ API call, and the property `$name` for displaying the issuer to your customer. For a more in-depth example, see [Example 4](https://github.com/mollie/mollie-api-php/blob/master/examples/04-ideal-payment.php)._
```
jimmyJames09z, line 98: Payme

jimmyJames09z (approved):

PR #26, opened 2016-06-15
disable Expect header

it-can:
The empty 'Expect' header, is saying to the server: please accept the body right now. This can speed up the request

More info: https://gist.github.com/perusio/1724301


mollierobbert:
Looks good it-can, thanks for your pull request.

We've merged this and published it as v1.6.5.

PR #27, opened 2016-06-27
Added cancellation functionality for Customer Subscriptions

Rkallenkoot:
Added rest_delete method along with a delete() method for convenience.
Added isValid & isCancelled method in Customer_Subscription class
Added example for cancelling a subscription (17-cancel-subscription.php)

Some help with the unit tests would be great :]

### DIFF
--- a/examples/16-recurring-subscription.php
+++ b/examples/16-recurring-subscription.php
@@ -10,6 +10,13 @@ try
 	 */
 	include "initialize.php";
 
+	/*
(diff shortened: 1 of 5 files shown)


mollierobbert:
Thanks Roelof, we've squash merged your changes into our master branch, with the addition of a unit test.

We've made the 'cancel' method an alias of the 'delete' method, to keep things simple.

Thanks again for your input.

PR #30, opened 2016-07-19
Improved flexibility with PEM files.

tobiastom:
We are deploying our application as a PHAR archive and curl will not be able to read inside this file. This pull request will make it possible to change the location of the PEM file, without re-implementing the `performHttpCall` function.

### DIFF
--- a/src/Mollie/API/Client.php
+++ b/src/Mollie/API/Client.php
@@ -282,7 +282,7 @@ class Mollie_API_Client
 	}
 
 	public function getPemPath() {
-		return realpath(dirname(__FILE__) . "/cacert.pem";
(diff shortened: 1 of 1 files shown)


mollierobbert:
Your change sounds sensible, thanks for your additions.

Instead of a public getter which a subclass would have to overwrite, we've implemented your idea as a public setter. This way you may call the method in the constructor of a subclass, or just call the method externally when instancing the API client.

These changes have made it into version v1.6.4.

Thanks again, Tobias.

tobiastom:
With that I don't even need to subclass it. Thank you very much for integrating it! 👍

PR #31, opened 2016-08-04
Fix cancel() parameter in 17-cancel-subscription.php

Rkallenkoot:
`cancel()` method in `Mollie_API_Resource_Customers_Subscriptions` only accepts the subscription_id string.

### DIFF
--- a/examples/17-cancel-subscription.php
+++ b/examples/17-cancel-subscription.php
@@ -29,7 +29,7 @@ try
 	 *
 	 * See: https://www.mollie.com/nl/docs/reference/subscriptions/cancel
 	 */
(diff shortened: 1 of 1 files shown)


juukie:
Lines 25 and 26 could be removed I guess? In addition, `$subscriptionId = isset($_GET['subscription_id']) ? $_GET['subscription_id'] : '';` could be simplified to `$subscriptionId = $_GET['subscription_id'];` because the link on `16-recurring-subscription.php
` will pass it through.

juukie:
Should lines 25 and 26 (retrieving subscription) be removed as well?

robin-mollie:
Has been amended in commit afb0467, including the subscription lookup being removed.

Thanks! :)

PR #33, opened 2016-08-08
Add isRecurring method

juukie:
So it's possible to do `$payment->isRecurring()` instead of `$payment->recurringType === \Mollie_API_Object_Payment::RECURRINGTYPE_FIRST`

### DIFF
--- a/src/Mollie/API/Object/Payment.php
+++ b/src/Mollie/API/Object/Payment.php
@@ -322,6 +322,16 @@ class Mollie_API_Object_Payment
 	}
 
 	/**
+	 * Is this payment recurring?
+	 *
+	 * @return bool
+	 */
+	public function isRecurring ()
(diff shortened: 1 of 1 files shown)


mollierobbert:
Good idea to add boolean methods, juukie. However, in this case, it may lead to confusion. We would actually only consider a payment to be 'recurring' when `recurringType=recurring`.

We've introduced `recurringType=first` as a way of setting up a regular payment that generates a mandate behind the scenes. The naming is confusing, but consistent with SEPA Direct Debit.

We've merged this, but changed the method name to `hasRecurringType`. Also, `hasRecurringTypeFirst` and `hasRecurringTypeRecurring` methods have been added.

juukie:
Makes sence @mollierobbert. Great solution. Thanks.

PR #36, opened 2016-09-09
Added cancel refund

happyDemon:


mollierobbert:
Thanks for your pull request, Maxim. It's already possible to cancel refunds via `$refunds_resource->delete($refund_id)`. It's a good idea to add an alias, but we'll keep it consistent with the way we've added the `cancel` alias to the Subscriptions resource. We've just added version 1.6.7 to support this.

mollierobbert:
https://github.com/mollie/mollie-api-php/commit/b6e73f4d237d279f619d4fc603a17b3b45a0966d

happyDemon:
Ah, sorry, it was just, I was browsing the online docs and it said it wasn't implemented (https://www.mollie.com/be/docs/reference/refunds/delete#example) in the PHP wrapper, so I just assumed it wasn't in the current codebase.

mollierobbert:
Good point :-). We've just updated the docs and clearly forgot to add the example there. Thanks for pointing that out, and sorry you had to go through the trouble of creating a pull request.

We'll update the example right away.

PR #21, opened 2016-03-15
Small fixes

RicardoJ:
* Added `version` field to composer.json for issue #20.
* Updated bundled cacert.pem file with latest changes.
* Updated library to 1.4.1.

PR #38, opened 2016-09-16
Issue/21 add customer update

Thijs-Riezebeek:
Also fixes issue [#21](https://github.com/mollie/laravel-mollie/issues/21) from the laravel-mollie module.


Daanvm:
Thanks for the PR.
Maybe it's nice to also add a new example?

PR #43, opened 2016-10-31
Add isPaidOut to the Payment object and bring the entire object under…

Thijs-Riezebeek:
* Make `getAmountRemaining` and `getAmountRefunded` always return a float.
* Add `isPaidOut`
* Add tests for the `Mollie_API_Object_Payment` class.


xewl:
Thanks for the commit, Thijs!

PR #54, opened 2016-12-28
Security fixes

ptondereau:
This PR provides security hardening in PHP code.

### DIFF
--- a/src/Mollie/API/Autoloader.php
+++ b/src/Mollie/API/Autoloader.php
@@ -39,7 +39,7 @@ class Mollie_API_Autoloader
 		if (strpos($class_name, "Mollie_") === 0)
 		{
 			$file_name = str_replace("_", "/", $class_name);
-			$file_name = realpath(dirname(__FILE__) . "/../../{$file_name}.php");
+			$file_name = realpath(__DIR__. "/../../{$file_name}.php");
 
 			if ($file_name !== false)
 			{
--- a/src/Mollie/API/Client.php
+++ b/src/Mollie/API/Client.php
@@ -183,6 +183,11 @@ class Mollie_API_Client
 	 */
 	protected $last_http_response_status_code;
 
+    /**
+     * @var Mollie_API_Resource_Profiles_APIKeys
+     */
+	public $profiles_apikeys;
+
 	/**
 	 * @throws Mollie_API_Exception_IncompatiblePlatform
 	 */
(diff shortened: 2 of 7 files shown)


src/Mollie/API/Client.php
```diff
@@ -396,7 +401,7 @@ public function performHttpCall ($http_method, $api_method, $http_body = NULL)
 			 */
 			$request_headers[] = "X-Mollie-Debug: old OpenSSL found";
 			curl_setopt($this->ch, CURLOPT_HTTPHEADER, $request_headers);
-			curl_setopt($this->ch, CURLOPT_SSL_VERIFYHOST, 0);
+			curl_setopt($this->ch, CURLOPT_SSL_VERIFYHOST, 2);
```
ptondereau: Prevent MITM attack

willemstuursma: Thanks. I guess it is time to get rid of this failover.

src/Mollie/API/Autoloader.php
```diff
@@ -39,7 +39,7 @@ public static function autoload ($class_name)
 		if (strpos($class_name, "Mollie_") === 0)
 		{
 			$file_name = str_replace("_", "/", $class_name);
-			$file_name = realpath(dirname(__FILE__) . "/../../{$file_name}.php");
```
willemstuursma: This won't work on some old PHP versions, sorry. We still need to support them.

src/Mollie/API/Client.php
```diff
@@ -216,7 +221,7 @@ public function __construct ()
 		$this->addVersionString($curl_version["ssl_version"]);
 
 		// The PEM path may be overwritten with setPemPath().
-		$this->pem_path = realpath(dirname(__FILE__) . "/cacert.pem");
```
willemstuursma: This won't work on some old PHP versions, sorry. We still need to support them.

willemstuursma:
Hi @ptondereau. I've removed the support for OpenSSL 0.9.7 in e0a4fd38f462b736462b613990cfaab31d131272. 

I'm afraid we won't be making the other changes, sorry.

Thanks for pointing this out and contributing to our module!

it-can:
But what about the strict checks? Seems reasonable?

ptondereau:
@it-can always do strict checks

it-can:
In this PR there are a few strict checks added... maybe a good addition to the Mollie PHP lib

PR #91, opened 2017-09-18
Update initialize.php

kloentje2:
Looks like the key from soneone from the dev department, I could be wrong but to be sure.


Smitsel:
This is just a test key so we would like it to stay.

As to the include -> require changes, please make one PR for it next time.
That would be a lot easier for us to check and merge.

kloentje2:
I will do!

PR #120, opened 2017-10-24
Patch php version in composer.json

Tjoosten:
Bump the php version in the composer.json file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ php:
 
 matrix:
   include:
-    - php: 5.3
+    - php: 5.4
       dist: precise
       sudo: true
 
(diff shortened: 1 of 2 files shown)


willemstuursma:
Thanks! We still want to support PHP 5.3, we can still run tests for that on CI. 

The reason for this is that our module is used a lot in Wordpress and WooCommerce.

willemstuursma:
@Smitsel can you bump the version to *PHP 5.3*.

Smitsel:
Will bump it to 5.3 instead of 5.4.
Opened new PR #123 

Thanks for putting this to our attention @Tjoosten !

PR #121, opened 2017-10-27
Implement a ConnectionError exception

ArondeParon:
# What am I looking at?

This pull request implements a custom exception for use cases in which a cURL error is returned. The exception contains the cURL error code and status message.

# But, why?

These errors were previously returned by the generic `Mollie_API_Exception`. The problem with this is that it's not very easy to distinguish between the different error types. For example, a cURL timeout would result in `error code 28` being returned, but one would have to parse the exception message to determine that this was the actual error.

With this pull request, handling different connection scenarios should be easier.

### DIFF
--- a/src/Mollie/API/Client.php
+++ b/src/Mollie/API/Client.php
@@ -392,14 +392,10 @@ class Mollie_API_Client
 
 		if (curl_errno($this->ch))
 		{
-			$message = "Unable to communicate with Mollie (".curl_errno($this->ch)."): " . curl_error($this->ch) . ".";
-
 			$this->closeTcpConnection();
 			
-			$exception = new Mollie_API_Exception_ConnectionError($message);
-			$exception->errorCode = curl_errno($this->ch);
-			$exception->errorMessage = curl_error($this->ch);
-			
+			$exception = Mollie_API_Exception_ConnectionError::fromCurlFailure($this->ch);
+
 			throw $exception;
 		}
 
--- a/src/Mollie/API/Client.php
+++ b/src/Mollie/API/Client.php
@@ -395,7 +395,12 @@ class Mollie_API_Client
 			$message = "Unable to communicate with Mollie (".curl_errno($this->ch)."): " . curl_error($this->ch) . ".";
 
 			$this->closeTcpConnection();
-			throw new Mollie_API_Exception($message);
+			
+			$exception = new Mollie_API_Exception_ConnectionError($message);
+			$exception->errorCode = curl_errno($this->ch);
+			$exception->errorMessage = curl_error($this->ch);
+			
+			throw $exception;
 		}
 
 		if (!function_exists("curl_reset"))
--- a/src/Mollie/API/Client.php
+++ b/src/Mollie/API/Client.php
@@ -393,10 +393,8 @@ class Mollie_API_Client
 		if (curl_errno($this->ch))
 		{
 			$this->closeTcpConnection();
-			
-			$exception = Mollie_API_Exception_ConnectionError::fromCurlFailure($this->ch);
 
-			throw $exception;
+			throw Mollie_API_Exception_ConnectionError::fromCurlFailure($this->ch);
 		}
 
 		if (!function_exists("curl_reset"))
--- a/src/Mollie/API/Client.php
+++ b/src/Mollie/API/Client.php
@@ -392,9 +392,15 @@ class Mollie_API_Client
 
 		if (curl_errno($this->ch))
 		{
+			$exception = Mollie_API_Exception_ConnectionError::fromCurlFailure($this->ch);
+
 			$this->closeTcpConnection();
 
-			throw Mollie_API_Exception_ConnectionError::fromCurlFailure($this->ch);
+			/*
+			 * We intentionally throw the exception after creating it and closing the connection because closing the
+			 * connection will reset the cull resource to null.
+			 */
+			throw $exception;
 		}
 
 		if (!function_exists("curl_reset"))
--- a/src/Mollie/API/Exception/ConnectionError.php
+++ b/src/Mollie/API/Exception/ConnectionError.php
@@ -1,0 +1,34 @@
+<?php
+/**
+ * Copyright (c) 2013, Mollie B.V.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE FOR ANY
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ *
+ * @license     Berkeley Software Distribution License (BSD-License 2) http://www.opensource.org/licenses/bsd-license.php
+ * @author      Mollie B.V. <info@mollie.com>
+ * @copyright   Mollie B.V.
+ * @link        https://www.mollie.com
+ */
+class Mollie_API_Exception_ConnectionError extends Mollie_API_Exception
+{
+}
--- a/src/Mollie/API/Exception/ConnectionError.php
+++ b/src/Mollie/API/Exception/ConnectionError.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Copyright (c) 2013, Mollie B.V.
  * All rights reserved.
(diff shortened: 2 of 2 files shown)


Thijs-Riezebeek:
Great idea! I'll review your changes.

src/Mollie/API/Exception/ConnectionError.php
```diff
@@ -0,0 +1,36 @@
...
+ * DAMAGE.
+ *
+ * @license     Berkeley Software Distribution License (BSD-License 2) http://www.opensource.org/licenses/bsd-license.php
+ * @author      Mollie B.V. <info@mollie.com>
+ * @copyright   Mollie B.V.
+ * @link        https://www.mollie.com
+ */
+class Mollie_API_Exception_ConnectionError extends Mollie_API_Exception
+{
+	public $errorCode;
```
Thijs-Riezebeek: Could you make these private and add getters?

The instantiation could then be done from a static method on the exception like so:

```php
// Docblock here
public static function fromCurlFailure($curl)
{
    $e = new self("Unable to communicate with Mollie (".curl_errno($curl)."): " . curl_error($curl) . ".";);

    $e->errorCode = curl_errno($curl);
    $e->errorMessage = curl_error($curl);

    return $e;
}
```

The exception message no longer needs to be generated from within the Client, the static method can determine what message to set.

src/Mollie/API/Client.php
```diff
@@ -395,7 +395,12 @@ public function performHttpCall ($http_method, $api_method, $http_body = NULL)
 			$message = "Unable to communicate with Mollie (".curl_errno($this->ch)."): " . curl_error($this->ch) . ".";
 
 			$this->closeTcpConnection();
-			throw new Mollie_API_Exception($message);
+			
+			$exception = new Mollie_API_Exception_ConnectionError($message);
```
Thijs-Riezebeek: `throw Mollie_API_Exception_ConnectionError::fromCurlFailure($this->ch);`

willemstuursma:
Thanks @Thijs-Riezebeek! I agree that this is a great idea, thanks @ArondeParon for starting the discussion.

I understand that you need to distinguish between connection and other errors. What kind of problems are you experiencing?

src/Mollie/API/Exception/ConnectionError.php
ArondeParon: Good idea, will do some refactoring for that.

src/Mollie/API/Exception/ConnectionError.php
```diff
@@ -0,0 +1,82 @@
...
+ * DAMAGE.
+ *
+ * @license     Berkeley Software Distribution License (BSD-License 2) http://www.opensource.org/licenses/bsd-license.php
+ * @author      Mollie B.V. <info@mollie.com>
+ * @copyright   Mollie B.V.
+ * @link        https://www.mollie.com
+ */
+class Mollie_API_Exception_ConnectionError extends Mollie_API_Exception
+{
+	private $curlErrorCode;
```
Thijs-Riezebeek: Could you add an `@var int` to the error code? I noticed you gave `getCurlErrorCode()` a `mixed` return type, but `curl_errno` is guaranteed to return an `int`.

src/Mollie/API/Exception/ConnectionError.php
```diff
@@ -0,0 +1,82 @@
...
+	private $curlErrorMessage;
```
Thijs-Riezebeek: Could you add an `@var string` to the error code? I noticed you gave `getCurlErrorMessage()` a `mixed` return type, but `curl_error` is guaranteed to return an `string`.

Thijs-Riezebeek:
@ArondeParon Sorry for making you go back again! I just had some last comments about the types used in the exception class.

ArondeParon:
@Thijs-Riezebeek no problem, done! 😄 

> I understand that you need to distinguish between connection and other errors. What kind of problems are you experiencing?

@willemstuursma this PR's is a 'scratch your own itch' thing 😉  We had an API exception occur that seemed to be caused by a cURL timeout with the Mollie API. In my payment handler, I'd like to be able to catch this scenario and do some proper error handling.

lode:
I've experienced timeout issues as well. In some case I re-instantiated the Mollie client for every call, after caching that most of the timeout issues were gone. Maybe a good thing to check as well.

Apart from that, I still get timeouts every now and then.

src/Mollie/API/Client.php
```diff
@@ -392,10 +392,11 @@ public function performHttpCall ($http_method, $api_method, $http_body = NULL)
 
 		if (curl_errno($this->ch))
 		{
-			$message = "Unable to communicate with Mollie (".curl_errno($this->ch)."): " . curl_error($this->ch) . ".";
-
 			$this->closeTcpConnection();
-			throw new Mollie_API_Exception($message);
+			
+			$exception = Mollie_API_Exception_ConnectionError::fromCurlFailure($this->ch);
```
Thijs-Riezebeek: Just throw in one go here (`throw Mollie_API...`).

src/Mollie/API/Exception/ConnectionError.php
```diff
@@ -0,0 +1,89 @@
...
+ * @license     Berkeley Software Distribution License (BSD-License 2) http://www.opensource.org/licenses/bsd-license.php
+ * @author      Mollie B.V. <info@mollie.com>
+ * @copyright   Mollie B.V.
+ * @link        https://www.mollie.com
+ */
+class Mollie_API_Exception_ConnectionError extends Mollie_API_Exception
+{
+    /**
+     * @var int
+     */
```
Thijs-Riezebeek: Is it me or is this indentation off a bit?

src/Mollie/API/Exception/ConnectionError.php
```diff
@@ -0,0 +1,89 @@
...
+    public function getCurlErrorCode()
+    {
+        return $this->curlErrorCode;
+    }
+
+    /**
+     * @param int $curlErrorCode
+     */
+    public function setCurlErrorCode($curlErrorCode)
```
Thijs-Riezebeek: There is no need for the setters, so it would be cleaner if they weren't there.

src/Mollie/API/Exception/ConnectionError.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+	private $curlErrorMessage;
+
+    /**
+     * @param resource $curl
+     * @return Mollie_API_Exception_ConnectionError
+     */
+    public static function fromCurlFailure($curl)
+    {
+        $e = new static("Unable to communicate with Mollie (".curl_errno($curl)."): " . curl_error($curl) . ")");
```
Thijs-Riezebeek: This exception message does not exactly match the exception message that was set previously. This is important since people might be using a regex to parse the Curl error numbers and strings from the exception and we don't want to break that.

The `")"` at the end should be `"."`, right?

Thijs-Riezebeek: @ArondeParon This was actually the reason I started another review round, I wanted to make sure the exception message was kept the same. You're probably not the only one who was struggling with getting more information out of this exception, so my guess is that a number of integrations retrieve it from the exception message.

ArondeParon: Good point, I'll look it. These are all valid points, but I need to reserve some spare time for this. Last few commits were a bit messy because I'm doing this in between tasks, so I want to get it right this time 😄

src/Mollie/API/Exception/ConnectionError.php
ArondeParon: Fixed! Was still using PSR-2 code style, causing indentation to be made with spaces instead of tabs.

Thijs-Riezebeek:
@ArondeParon Thanks for your changes Aron! This should definitely help anyone using this package to perform better error handling, logging and debugging.

We will merge this PR very soon!

src/Mollie/API/Client.php
```diff
@@ -392,10 +392,9 @@ public function performHttpCall ($http_method, $api_method, $http_body = NULL)
 
 		if (curl_errno($this->ch))
 		{
-			$message = "Unable to communicate with Mollie (".curl_errno($this->ch)."): " . curl_error($this->ch) . ".";
-
 			$this->closeTcpConnection();
-			throw new Mollie_API_Exception($message);
+
+			throw Mollie_API_Exception_ConnectionError::fromCurlFailure($this->ch);
```
lode: Does this work if `closeTcpConnection` cleans up the curl handler already?

willemstuursma: No it doesn't, it will set `$ch` to `null`. @Thijs-Riezebeek

Thijs-Riezebeek: @ArondeParon That's my bad, this won't work indeed. It should be 

```
$e = Mollie_API_Exception_ConnectionError::fromCurlFailure($this->ch);

$this->closeTcpConnection();

throw $e;
```

like you had before. Somehow I didn't see that calling `closeTcpConnection()` first would result in problems.

Thijs-Riezebeek: @ArondeParon That's my bad, this won't work indeed. It should be 

```
$e = Mollie_API_Exception_ConnectionError::fromCurlFailure($this->ch);

$this->closeTcpConnection();

throw $e;
```

like you had before. Somehow I didn't see that calling `closeTcpConnection()` first would result in problems.

ArondeParon:
No problem! Adjusted the routine to create an Exception instance *before* closing the connection, which should solve this issue.

lode:
> We've been having trouble figuring out where these timeouts are coming from. If you have any additional info on these (timestamps, server locations and such), could you share it?

Sure @mollierobbert! I can easily give a list of timestamps, transaction ids, the ip of the webhook call. What more would be helpful? The location of our server?

mollierobbert:
Yes, that would be awesome @lode. Thanks for helping us out.

We specifically need timestamps and the server context - location, hosting provider... Something like that.

Could you send them over to robbert@mollie.com?

willemstuursma:
@lode we have received your list.

What does your setup roughly look like, e.g.:

1. You are creating transactions, and use a new `Mollie_API_Client` for every API call
2. You are creating transactions in bulk, and use a new `Mollie_API_Client` for every API call
3. You are creating transactions in bulk, and reusing one `Mollie_API_Client` object for all calls
4. A different usage of `Mollie_API_Client`.

willemstuursma:
@ArondeParon it would be helpful if you could elaborate on your usage too?

ArondeParon:
Will do. I'll need some time to gather this info, but you can expect it somewhere next week, or sooner if I can squeeze in a moment this weekend 😄

lode:
> What does your setup roughly look like

Mainly 1 and 3.

For migrations we use 3. This is an area we saw a lot of timeouts before, then switched to caching the client, and saw a huge drop in timeouts. They didn't disappear though.

For day-to-day usage we use 1. Per request of an end-user we create a new client. Mainly because it is request-based. If an end-user ever would create multiple transactions, the client would have been cached.

I don't see anything very special in constructing the client btw. It is just creating some objects of classes and setting a few properties. Is something hidden further down which would cause timeouts?

ArondeParon:
@willemstuursma regarding usage: we're using `Mollie_API_Client` on a per-request basis, so that would be scenario 1.

Would timestamps and additional info still help @mollierobbert ?

willemstuursma:
> Is something hidden further down which would cause timeouts?

The `Mollie_API_Client` object holds a cURL handle which holds an open socket.

Thanks @ArondeParon and @lode for sharing your usage of our client.

We're looking into this together with our hosting supplier.

PR #125, opened 2017-11-08
Retry on connection error

willemstuursma:
This retries the API call if the connection cannot be established - e.g. if the socket cannot be opened for some reason. So, only if no packets were received by the Mollie API.


Smitsel:
LGTM! Thanks

PR #126, opened 2017-11-09
Update cacert.pem + cleanup code + stricter checks

it-can:


martijnpolder:
I verified the checksum of the cacert.pem file - it's valid.

it-can:
Downloaded it from https://curl.haxx.se/docs/caextract.html

Smitsel:
Thanks!

PR #129, opened 2017-12-13
Cancel payment

insit:
Created this method to keep in line with cancelling refunds.


arendvw:
Just curious: What problem does this pull request solve?

insit:
I was exploring the API docs and saw that there was no direct cancel method
on payments for the PHP wrapper. There is one for refunds, I know it is
just an alias and I can use the delete call but to keep it in line with
refunds I made this PR.

Kind regards and happy holidays!

Tom

Op 13 dec. 2017 15:27 schreef "arendvw" <notifications@github.com>:

Smitsel:
I can see why this is nice to have.
Since we also have this alias method on our Subscriptions resource.

Thanks for contributing!

PR #132, opened 2018-01-02
Make it so we expect 303 client behaviour when redirecting fixes #119

### LINKED ISSUE #119: Add HTTP code to Location headers
Prevent clients from accidentally using 307 redirects which will break some (but not all) payment methods / issuers. Correct code is 303. https://stackoverflow.com/a/4764473/6632554

Smitsel:
We want to make unambiguously clear which kind of http redirect reaction we expect for clients using our examples. So we add http response code 303 to our header locations.

We do this to make sure that when our example is posted upon and then redirects (for example when making a post form with a payment method) we do a GET on the location field-value even when we use a POST as the original request method.

There have been cases that it got treated as 307 and copied the original request method for the redirect. Which can result in failing payments.

See the note on the 302 response: 
https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.3

### DIFF
--- a/examples/01-new-payment.php
+++ b/examples/01-new-payment.php
@@ -48,9 +48,10 @@ try
 	 */
 	database_write($order_id, $payment->status);
 
-	/*
-	 * Send the customer off to complete the payment.
-	 */
+    /*
+     * Send the customer off to complete the payment.
+     * This request should always be a GET, thus we enforce 303 http response code
+     */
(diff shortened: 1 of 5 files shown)


Smitsel:
The redirect should always be a GET request. So we enforce 303.

willemstuursma:
👍

PR #133, opened 2018-01-03
Replace IBAN in example that doesent pass checks

Smitsel:
The IBAN inside the 22-create-mandate-subscription.php does not pass our API's IBAN checks.
I generated a non-existing IBAN which does pass these checks and added it to our example.

Now it runs without error when used on a test API key.


willemstuursma:
Thanks.

PR #135, opened 2018-01-11
Create example delete customer

rconjour:
Example how to delete a customer by api


rconjour:
Hi,
I've added the comment and renamed the file. But an output is not possible (your api returns 204 - no output).

Smitsel:
@rconjour thats fast!
Just the echo after the delete now and im very happy :)

Smitsel:
Thanks @rconjour. And congrats on your first contribution. Much appreciated!

PR #139, opened 2018-01-27
Add canBeCancelled property to Mollie_API_Object_Payment

insit:
Added this property to the Mollie_API_Object_Payment class to keep it in line with the docs.

### DIFF
--- a/src/Mollie/API/Object/Payment.php
+++ b/src/Mollie/API/Object/Payment.php
@@ -258,6 +258,13 @@ class Mollie_API_Object_Payment
 	 */
 	public $links;
 
+    /**
+     * Whether or not the payment can be cancelled.
+     *
+     * @var bool|null
+     */
+	public $canBeCancelled;
+
 	/**
 	 * Is this payment cancelled?
 	 *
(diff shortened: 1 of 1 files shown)


src/Mollie/API/Object/Payment.php
```diff
@@ -258,6 +258,13 @@ class Mollie_API_Object_Payment
 	 */
 	public $links;
 
+    /**
+     * Whether or not the payment can be cancelled.
+     *
+     * @var bool|null
+     */
```
Smitsel: Thanks for the PR, one small change request.
Could u make sure the indenting is right on this new docblock?

insit: Done :)

Smitsel:
Thanks again! We love consistency ;)

insit:
No problem at all! Could you maybe increase the version number in order for Composer to fetch a new version?

Kind regards!!

willemstuursma:
We'll release a new version tomorrow.

insit:
Thanks for releasing a new version. We are a very happy customer!

Kind regards!

Tom

PR #151, opened 2018-03-28
Remove stasuses and use the _links object for the helper methods

Smitsel:


src/Mollie/API/Resources/Payment.php
```diff
@@ -285,34 +292,31 @@ public function isPaid()
...
-     * Note: When a payment is refunded or charged back, the status 'refunded'/'charged_back' will
-     * overwrite the 'paidout' status.
-     *
      * @return bool
      */
-    public function isPaidOut()
+    public function hasSettlement()
     {
-        return $this->status === PaymentStatus::STATUS_PAIDOUT;
+        return !empty($this->settlementAmount);
```
Daanvm: This doesn't work, because the settlement amount is also returned when the payment is not settled yet. You have to look at the `settlementId` (which currently doesn't exist yet in the v2 API).

PR #154, opened 2018-03-29
Add methods resource

shirshir:


tests/Mollie/API/Endpoints/MethodEndpointTest.php
```diff
@@ -0,0 +1,240 @@
...
+            )
+        );
+
+        $idealMethod = $this->apiClient->methods->get('ideal');
+
+        $this->assertInstanceOf(Method::class, $idealMethod);
+        $this->assertEquals('ideal', $idealMethod->id);
+        $this->assertEquals('iDEAL', $idealMethod->description);
+
+        $amount = new Stdclass();
```
Daanvm: This is not an amount, and you're not using this variable anyway.

shirshir: Fixed

PR #155, opened 2018-04-05
2.0 cursor collections

Smitsel:


src/Mollie/API/Resources/ResourceFactory.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    /**
+     * Create resource object from Api result
+     *
+     * @param object $apiResult
+     * @param BaseResource $resource
+     *
+     * @return BaseResource
+     */
+    public static function createFromApiResult($apiResult, BaseResource $resource)
```
Daanvm: Can you add a simple unit test for this method?

src/Mollie/API/Resources/CursorCollection.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace Mollie\Api\Resources;
+
+use Mollie\Api\MollieApiClient;
+
+abstract class CursorCollection extends BaseCollection
```
Daanvm: The refunds collection should also extend from this class.

src/Mollie/API/Resources/ResourceFactory.php
Smitsel: Done.

src/Mollie/API/Resources/CursorCollection.php
Smitsel: Done!

PR #171, opened 2018-05-07
2.0 Add support for the mandate endpoint

Smitsel:


src/Endpoints/SubscriptionEndpoint.php
```diff
@@ -82,14 +82,14 @@ public function listFor(Customer $customer, $from = null, $limit = null, array $
 
     /**
      * @param Customer $customer
-     * @param string $subscriptionId
+     * @param string $mandateId
      *
      * @return object
      */
-    public function cancelFor(Customer $customer, $subscriptionId)
+    public function cancelFor(Customer $customer, $mandateId)
```
willemstuursma: Is this subscriptions or mandates?

willemstuursma:
Ok, small stuff.

src/Resources/Mandate.php
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * @var object
+     */
+    public $_links;
+
+    /**
+     * @return bool
+     */
+    public function isValid ()
```
mvdpanne: I'm not sure if this is the preferred style (the space between function name and parentheses).

tests/Mollie/API/Endpoints/MandateEndpointTest.php
```diff
@@ -0,0 +1,251 @@
+<?php
```
mvdpanne: I take it the `RevokeMandate` test was not added because it should return nothing, thus not being worth adding?

Smitsel: Yes correct, it would test absolutely nothing.

PR #172, opened 2018-05-07
2.0 Customer payments and examples

Smitsel:


examples/12-new-customer-payment.php
```diff
@@ -0,0 +1,69 @@
...
+     * See: https://www.mollie.com/en/docs/reference/customers/create-payment
+     */
+    $payment = $customer->createPayment(array(
+        "amount"       => [
+            "value" => "10.00",
+            "currency" => "EUR"
+        ],
+        "description"  => "My first Customer payment",
+        "redirectUrl"  => "{$protocol}://{$hostname}{$path}/03-return-page.php?order_id={$order_id}",
+        "webhookUrl"   => "{$protocol}://{$hostname}{$path}/02-webhook-verification.php"
```
willemstuursma: Add sequenceType to show that you can start a recurring payment.

Smitsel: I could do that, but wouldn't it be a duplicate of:
{{examples/14-recurring-first-payment.php}}

willemstuursma: OK, then maybe add some comments here on why adding a customerId is great.

PR #176, opened 2018-05-09
Update README.md

toonvandeputte:
Clarified that the client version corresponds to the API version


Smitsel:
Thanks @toonvandeputte !

PR #177, opened 2018-05-09
Make settlements client aware and add chargeback pagination

Smitsel:


willemstuursma:
This one is also used for the top level resource right?

Smitsel:
Yes for toplevel, and also $payment->chargebacks() and $settlement->chargebacks()

willemstuursma:
It looks fine but I have a feeling of code duplication. Can you clean that up by introducing a trait or an abstract class?

Smitsel:
Cleaned it up by expanding the abstract CursorCollection.
Great suggestion!

src/Resources/CursorCollection.php
```diff
@@ -14,26 +14,63 @@
     /**
      * @param MollieApiClient $client
      * @param int $count
-     * @param object $_links
+     * @param object[] $_links
      */
-    public function __construct(MollieApiClient $client, $count, $_links)
+    final public function __construct(MollieApiClient $client, $count, $_links)
```
willemstuursma: Make it protected, it's only instanciated by this class right?

src/Resources/CursorCollection.php
```diff
@@ -14,26 +14,63 @@
...
 
         $this->client = $client;
     }
 
+    /**
+     * Return the resource object
+     *
+     * @return BaseResource
+     */
+    abstract protected function getResourceObject();
```
willemstuursma: Maybe better call this `createResourceObject` because it has to create new ones.

willemstuursma:
This looks great.

src/Resources/CursorCollection.php
Smitsel: It extends from BaseCollection (the one without pagination) and ArrayObject.

PR #181, opened 2018-05-15
Some example improvements

ndijkstra:


Smitsel:
Thanks for this @ndijkstra

PR #188, opened 2018-05-31
2.0 Add profile endpoint to client

Smitsel:


src/Resources/Profile.php
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * @var object
+     */
+    public $review;
+
+    /**
+     * UTC datetime the profile was created in ISO-8601 format.
+     *
+     * @example "2013-12-25T10:30:54+00:00"
+     * @var string|null
```
shirshir: `createdAt` is never `null`

Smitsel:
Changed.

src/Resources/Profile.php
Smitsel: √

Issue #193, opened 2018-06-04
Delete mandate gives 204 exception


esprange:
if the revokeMandate method of customers is used, an exception is raised because of an empty body return. I used Postman to similate this and it returns indeed a 204 status. The work-around is a catching this exception, but in fact it should not be an exception at all...

willemstuursma:
Hi @esprange. Is this in the v1 or v2 version of this client?

esprange:
I am using v2.

esprange:
And Willem, I found the issue while in testmode. But I don't think that should matter.

Smitsel:
Hi @esprange,

We've added a check for the 204 response code to the parsing of the body.
So we won't throw an exception when the body is empty and the code is 204.

So you will now get an empty response. Since this is in line with the empty response from the api.
So you can use a try catch around the revoke, when it succeeds you can be sure that it is deleted.

PR #197, opened 2018-06-05
Revoke customer mandate no exception and helper function

Smitsel:
As noticed in #193 

When doing a delete call our API handles this by giving a 204 No Content response.
Which is good, but our API client should not give an exception.

So i decided to return null, an empty response.
I thought about return a true boolean as some kind of confirmation towards the client user.
But i thought it should be in line with the API so i didn't, and handle this as a succeed or throw.


willemstuursma:
Thanks!

PR #199, opened 2018-06-13
Improve return types and make the available endpoints explicit

Smitsel:
Changed the return types to the resource object classes.
And made the endpoint calls explicit, since for instance a /methods call only has get and list.
So when calling ``$mollie->methods->`` my IDE does not suggest create or delete.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -8,7 +8,7 @@
         <directory>tests/</directory>
     </testsuite>
     <logging>
-        <log type="junit" target="build/logs/junit-unittests.xml" logIncompleteSkipped="false"/>
+        <log type="junit" target="build/logs/junit-unittests.xml"/>
     </logging>
     <filter>
         <whitelist processUncoveredFilesFromWhitelist="true">
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -8,7 +8,7 @@
         <directory>tests/</directory>
     </testsuite>
     <logging>
-        <log type="junit" target="build/logs/junit-unittests.xml"/>
+        <log type="junit" target="build/logs/junit-unittests.xml" logIncompleteSkipped="false"/>
     </logging>
     <filter>
         <whitelist processUncoveredFilesFromWhitelist="true">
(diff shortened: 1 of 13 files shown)


src/Endpoints/EndpointAbstract.php
```diff
@@ -55,18 +55,23 @@ private function buildQueryString(array $filters)
...
      * @param string|null|resource|StreamInterface $body
      * @param array $filters
-     * @return object
+     * @return BaseResource
      * @throws ApiException
      */
-    private function rest_create($restResource, $body, array $filters)
+    protected function rest_create($body, array $filters)
     {
+        $encoded = json_encode($body);
```
willemstuursma: There's a Guzzle one that you can use `\GuzzleHttp\json_encode()`

phpunit.xml
```diff
@@ -8,7 +8,7 @@
         <directory>tests/</directory>
     </testsuite>
     <logging>
-        <log type="junit" target="build/logs/junit-unittests.xml" logIncompleteSkipped="false"/>
+        <log type="junit" target="build/logs/junit-unittests.xml"/>
```
willemstuursma: Unrelated change?

willemstuursma:
Minor stuff.

phpunit.xml
Smitsel: Yes, this flag did not exist for me.
But ill remove the change from this scope.

src/Endpoints/EndpointAbstract.php
Smitsel: √

Smitsel:
Thanks, i've updated your feedback 👍

src/Endpoints/EndpointAbstract.php
```diff
@@ -55,18 +55,23 @@ private function buildQueryString(array $filters)
...
-     * @return object
+     * @return BaseResource
      * @throws ApiException
      */
-    private function rest_create($restResource, $body, array $filters)
+    protected function rest_create($body, array $filters)
     {
+        $encoded = \GuzzleHttp\json_encode($body);
+
+        if (json_last_error() !== JSON_ERROR_NONE) {
```
willemstuursma: @Smitsel this Guzzle function already throws on failure.

PR #205, opened 2018-06-19
Update link to API keys page

lvgunst:
We've moved the API keys to a dedicated developers page in the dashboard. Only old merchants will still see a link to the API keys on the profiles page.


willemstuursma:
Do you also want to change it in the `v1-develop` branch?

PR #208, opened 2018-06-21
Add giropay & eps support for V2

mvdpanne:
While I was at it, figured it wouldn't hurt to alphabetically sort the payment method constants. Good to note is that all other payment methods have a dedicated link, but for eps & giropay it isn't present yet.

Furthermore, the following links for the payment methods do not work:
- https://www.mollie.com/directdebit --> redirects to login page.
- https://www.mollie.com/giftcards --> 404.


mvdpanne:
Giftcards payment link has now been fixed, I've asked Tjeerd about the DirectDebit link.

Milda has let me know that the pages for Giropay and eps will probably be live tomorrow. Afterwards we can fix those links.

willemstuursma:
I think the directdebit prefix is used for some internal dashboard pages.

mvdpanne:
After talking to Tjeerd, it seems like the best option would be to just change them all to `mollie.com/en/payments/[payment-method]`, instead of relying on redirects.

shirshir:
Yes, I agree.

davdebcom:
Don't forget to merge this, as it's now launched. I'm using the branch atm.

willemstuursma:
Thanks for the reminder @davdebcom.

PR #213, opened 2018-06-26
GuzzleException never thrown in block

davdebcom:
GuzzleException never thrown in block, shouldn't this be: \InvalidArgumentException


willemstuursma:
It would throw a `GuzzleException` when there is an SSL error, connection error, timeout etc.

PR #219, opened 2018-06-27
Add helper function for active mandate. Fixes #218

### LINKED ISSUE #218: Feature request: hasActiveMandate helper method
## Specifications

  - API Version: 1 and/or 2

## Describe the issue
Manually implementing the check whether a customer has an active mandate is a bit cumbersome. As far as I know there is no helper method for this.

How about an `$customer->hasActiveMandate()` helper?

Smitsel:


ndijkstra:
Isn't it better to use "hasValidMandate" ?

sandervanhooft:
I had the same thoughts, but semantically a mandate could be valid yet inactive.

sandervanhooft:
Yet, currently a `$mandate->isValid()` helper exists on the Mandate resource. So if consistency outweighs semantics I have to agree with @ndijkstra.

Smitsel:
Yes, i agree with @ndijkstra and changed it.

PR #229, opened 2018-07-17
#23 Isolate dependencies in build

willemstuursma:
This adds a step where the zip file that is used for releases is special - Mollie's dependencies have been isolated using [humbug/php-scoper](https://github.com/humbug/php-scoper).

An example release has been created using this method: 

[mollie-api-php.zip](https://github.com/mollie/mollie-api-php/files/2202345/mollie-api-php.zip)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,25 @@
+#
+# This automatically creates a fresh build for distribution with other modules or for getting started with Mollie
+# without composer.
+#
+mollie-api-php.zip:
+	#
+	# First, install all dependencies. Then prefix everything with humbug/php-scoper. Finally, we should dump the
+	# autoloader again to update the autoloader with the new classnames.
+	#
+	composer install --no-dev --no-scripts --no-suggest
+	$(shell composer global config bin-dir --absolute)/php-scoper add-prefix --force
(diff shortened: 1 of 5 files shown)


willemstuursma:
@davdebcom can you test this file?

Makefile
```diff
@@ -0,0 +1,25 @@
...
+	#
+	composer install --no-dev --no-scripts --no-suggest
+	$(shell composer global config bin-dir --absolute)/php-scoper add-prefix --force
+	composer dump-autoload --working-dir build --classmap-authoritative
+
+	#
+	# Now move the autoload files. We have to use the scoper one to load the aliasses but we want to load the normal
+	# filename. Flip them around.
+	#
+	mv build/vendor/autoload.php build/vendor/composer-autoload.php
```
Daanvm: The contents of `autoload.php` (previously called `scoper-autoload.php`) doesn't look very interesting. Do we really need it?

willemstuursma: Yes, else you cannot inject a `GuzzleClient` as it will not match the typehint anymore.

willemstuursma: Well, users of this zipfile wouldn't really do that right.

PR #230, opened 2018-07-18
Abstract the Guzzle dependency to a HttpClient implementation

Smitsel:
This will be a BC breaking change. Since the install will now require 2 steps.
Installing the SDK and also installing a HTTP client implementation.

This will not matter for our ZIP file as we can require the Guzzle 6 adapter in the steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,13 +19,17 @@ To use the Mollie API client, the following things are required:
 
 ## Composer Installation ##
 
-By far the easiest way to install the Mollie API client is to require it with [Composer](http://getcomposer.org/doc/00-intro.md).
+By far the easiest way to install the Mollie API client is to require it with [Composer](http://getcomposer.org/doc/00-intro.md).  
 
-    $ composer require mollie/mollie-api-php:^2.0
+The Mollie client depends on a ``php-http/client-implementation`` so make sure to install a package providing that.
+We suggest using ``php-http/guzzle6-adapter``.
+
+    $ composer require mollie/mollie-api-php:^2.0 php-http/guzzle6-adapter:^1.0
 
     {
         "require": {
-            "mollie/mollie-api-php": "^2.0"
+            "mollie/mollie-api-php": "^2.0",
+            "php-http/guzzle6-adapter": "^1.0"
         }
     }
 
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ To use the Mollie API client, the following things are required:
 + Follow [a few steps](https://www.mollie.com/dashboard/?modal=onboarding) to enable payment methods in live mode, and let us handle the rest.
 + PHP >= 5.6
 + Up-to-date OpenSSL (or other SSL/TLS toolkit)
++ An HTTP client which implements the ``php-http/client-implementation``. You can find the available implementations [here](https://packagist.org/providers/php-http/client-implementation). We recommend ``php-http/guzzle6-adapter``.
 
 ## Composer Installation ##
 
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The Mollie client depends on a ``php-http/client-implementation`` so make sure t
 We suggest using ``php-http/guzzle6-adapter``.
 
 ```console
-$ composer require mollie/mollie-api-php:^2.0 php-http/guzzle6-adapter:^1.0
+$ composer require mollie/mollie-api-php:^2.1 php-http/guzzle6-adapter:^1.0
 ```
 
 The version of the API client corresponds to the version of the API it implements. Check the [notes on migration](https://docs.mollie.com/migrating-v1-to-v2) to see what changes you need to make if you want to start using a newer API version.
--- a/composer.json
+++ b/composer.json
@@ -47,11 +47,18 @@
     "ext-curl": "*",
     "ext-json": "*",
     "ext-openssl": "*",
-    "guzzlehttp/guzzle": "^6.3"
+    "psr/http-message": "^1.0",
+    "php-http/client-implementation": "^1.0",
+    "php-http/discovery": "^1.0",
+    "php-http/message-factory": "^1.0"
+
   },
   "require-dev": {
+    "eloquent/liberator": "^2.0",
     "phpunit/phpunit": "^5.7|^6.5|^7.1",
-    "eloquent/liberator": "^2.0"
+    "php-http/mock-client": "^1.1",
+    "php-http/message": "^1.6",
+    "guzzlehttp/psr7": "^1.0"
   },
   "autoload": {
     "psr-4": {
--- a/src/Endpoints/EndpointAbstract.php
+++ b/src/Endpoints/EndpointAbstract.php
@@ -63,7 +63,7 @@ abstract class EndpointAbstract
     protected function rest_create($body, array $filters)
     {
         try {
-            $encoded = \GuzzleHttp\json_encode($body);
+            $encoded = json_encode($body);
         } catch (\InvalidArgumentException $e) {
             throw new ApiException("Error encoding parameters into JSON: '" . $e->getMessage() . "'.");
         }
--- a/src/MollieApiClient.php
+++ b/src/MollieApiClient.php
@@ -26,7 +26,7 @@ class MollieApiClient
     /**
      * Version of our client.
      */
-    const CLIENT_VERSION = "2.0.5";
+    const CLIENT_VERSION = "2.0.6";
 
     /**
      * Endpoint of the remote API.
--- a/src/MollieApiClient.php
+++ b/src/MollieApiClient.php
@@ -2,10 +2,11 @@
 
 namespace Mollie\Api;
 
-use GuzzleHttp\Client;
-use GuzzleHttp\ClientInterface;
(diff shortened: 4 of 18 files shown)


willemstuursma:
@Smitsel can you resolve the merge conflicts?

src/Endpoints/EndpointAbstract.php
```diff
@@ -63,7 +63,7 @@ private function buildQueryString(array $filters)
     protected function rest_create($body, array $filters)
     {
         try {
-            $encoded = \GuzzleHttp\json_encode($body);
+            $encoded = json_encode($body);
```
willemstuursma: Now we need to check if an error occurred again, not with try ... catch but with `json_error_last`

src/MollieApiClient.php
```diff
@@ -149,13 +155,15 @@ class MollieApiClient
...
+     * @param HttpClient $httpClient
+     * @param RequestFactory $requestFactory
      *
      * @throws IncompatiblePlatform
      */
-    public function __construct(ClientInterface $httpClient = null)
+    public function __construct(HttpClient $httpClient = null, RequestFactory $requestFactory = null)
     {
-        $this->httpClient = $httpClient ? $httpClient : new Client();
+        $this->httpClient = $httpClient ? $httpClient : HttpClientDiscovery::find();
```
willemstuursma: Let's wrap this in a nice exception if we can't find it (one of our exceptions) that has some actionable message so that the merchant knows what to do.

src/MollieApiClient.php
```diff
@@ -397,6 +404,6 @@ public function __sleep()
      */
     public function __wakeup()
     {
-        $this->__construct();
+        $this->__construct($this->httpClient);
```
willemstuursma: $this->httpClient will be null after waking up, just pass null.

tests/Mollie/API/MollieApiClientTest.php
```diff
@@ -120,7 +111,7 @@ public function testCanBeSerializedAndUnserialized()
         $client_copy = Liberator::liberate(unserialize($serialized));
 
         $this->assertEmpty($client_copy->apiKey, "API key should not have been remembered");
-        $this->assertInstanceOf(ClientInterface::class, $client_copy->httpClient, "A Guzzle client should have been set.");
+        $this->assertInstanceOf(HttpClient::class, $client_copy->httpClient, "A Guzzle client should have been set.");
```
willemstuursma: Please update the assertion message

composer.json
```diff
@@ -47,11 +47,18 @@
...
+    "php-http/message-factory": "^1.0"
+
   },
   "require-dev": {
+    "eloquent/liberator": "^2.0",
     "phpunit/phpunit": "^5.7|^6.5|^7.1",
-    "eloquent/liberator": "^2.0"
+    "php-http/mock-client": "^1.1",
+    "php-http/message": "^1.6",
+    "guzzlehttp/psr7": "^1.0"
```
willemstuursma: We don't need guzzle for dev right, we have the mock client

Smitsel: The mock client does not contain any ways to create a Request or Response object.
Thats why we need the psr7 package.

willemstuursma: oh sorry !

src/MollieApiClient.php
Smitsel: The httpClient is now also added to sleep. This was however done to make sure the serialisation test works. So we could leave it out but then we can't have that test.

This is because the discovery will throw an exception because we don't have any discoverable implementations installed.

README.md
```diff
@@ -19,13 +19,17 @@ To use the Mollie API client, the following things are required:
...
-By far the easiest way to install the Mollie API client is to require it with [Composer](http://getcomposer.org/doc/00-intro.md).
+By far the easiest way to install the Mollie API client is to require it with [Composer](http://getcomposer.org/doc/00-intro.md).  
 
-    $ composer require mollie/mollie-api-php:^2.0
+The Mollie client depends on a ``php-http/client-implementation`` so make sure to install a package providing that.
+We suggest using ``php-http/guzzle6-adapter``.
+
+    $ composer require mollie/mollie-api-php:^2.0 php-http/guzzle6-adapter:^1.0
 
     {
```
willemstuursma: Let's remove this blurp of JSON and just keep the shell command. Any way to add syntax highlighting to it?

willemstuursma:
The API changes only very narrowly. I feel we can get away with releasing a v2.1 version.

@shirshir @ndijkstra what do you think?

src/MollieApiClient.php
```diff
@@ -162,8 +163,13 @@ class MollieApiClient
      */
     public function __construct(HttpClient $httpClient = null, RequestFactory $requestFactory = null)
     {
-        $this->httpClient = $httpClient ? $httpClient : HttpClientDiscovery::find();
-        $this->requestFactory = $requestFactory ? $requestFactory : MessageFactoryDiscovery::find();
+
+        try {
+            $this->httpClient = $httpClient ? $httpClient : HttpClientDiscovery::find();
+            $this->requestFactory = $requestFactory ? $requestFactory : MessageFactoryDiscovery::find();
+        } catch (Exception $e) {
```
willemstuursma: @Smitsel any more specific exceptions available or this is the one?

src/MollieApiClient.php
```diff
@@ -162,8 +163,13 @@ class MollieApiClient
...
+            throw new IncompatiblePlatform('We could not find any http-client-implementation. Install a package providing "php-http/client-implementation". Example: "php-http/guzzle6-adapter".');
```
willemstuursma: `Example: "composer install php-http/guzzle6-adapter".`

tests/Mollie/API/Endpoints/PaymentEndpointTest.php
```diff
@@ -180,7 +180,7 @@ public function testGetPayment()
             )
         );
 
-        $payment = $this->apiClient->payments->get("tr_44aKxzEbr8", ["testmode" => true]);
```
willemstuursma: Please put this back!

tests/Mollie/API/MollieApiClientTest.php
```diff
@@ -111,7 +111,7 @@ public function testCanBeSerializedAndUnserialized()
         $client_copy = Liberator::liberate(unserialize($serialized));
 
         $this->assertEmpty($client_copy->apiKey, "API key should not have been remembered");
-        $this->assertInstanceOf(HttpClient::class, $client_copy->httpClient, "A Guzzle client should have been set.");
+        $this->assertInstanceOf(HttpClient::class, $client_copy->httpClient, "A Http client implementation should have been set.");
```
willemstuursma: _An Http_

willemstuursma:
This looks fine to me. Only change the readme to use `^2.1` as the composer version.

src/MollieApiClient.php
willemstuursma: In the test, make the httpmock client discoverable.

willemstuursma: We can't serialize the client.

barryvdh:
It might be a small BC break, but existing projects will run into this when updating their composer packages. Eg. not just this one, but also your magento2 packages and peoples own projects. They will get a warning from Composer, but it may not be very clear for everyone what they need to do exactly..

tests/Mollie/API/Endpoints/BaseEndpointTest.php
```diff
@@ -2,62 +2,46 @@
...
+     */
+    protected $messageFactory;
 
     /**
      * @var MollieApiClient
      */
     protected $apiClient;
 
-    protected function mockApiCall(Request $expectedRequest, Response $response)
+    protected function setUp()
```
willemstuursma: Sorry, this method should now be removed.

Smitsel:
The addition of requiring guzzle in building te zip is good.
But now the zip does not work because the scoped Guzzle cannot be discovered.

sandervanhooft:
Barryvdh wrote
> It might be a small BC break, but existing projects will run into this when updating their composer packages. Eg. not just this one, but also your magento2 packages and peoples own projects. They will get a warning from Composer, but it may not be very clear for everyone what they need to do exactly..

There's no such thing as a "small breaking change". 

As a consumer of this package, I'd request to put this on the backlog for the next major release.

It's not a must-have for 99% (100%?) of current implementations, just an architecture consideration. The practical value is low. Implementing it in a minor release by hacking it into the ZIP seems risky, imho it's not worth it.

(Don't get me wrong, I love good architecture.)

sandervanhooft:
@Smitsel Perhaps the Orders & Shipments features (non-BC) along with this PR would make up a nice next major release.

Smitsel:
Yeah it might be, but we're still a bit sceptical if we should proceed with this PR.
I'm leaving it up to @willemstuursma to make the final decision.

sandervanhooft:
I share your concerns. Have not been convinced of HttpClient abstraction in the past; Guzzle is still leading.

Smitsel:
This PR is getting quite old now and as we did not encounter problems with this anymore i'm rejecting this until the need arises again.

Issue #235, opened 2018-07-24
$payment->getRefund(...) missing


sandervanhooft:
## Specifications

  - API Version: v2.0
  - Mollie client: v2.0.8

## Describe the issue
Working on the API docs, discovered there's no `$client->refunds->get('refund_id_here')` method. Makes it hard to retrieve or cancel a single refund.

Not worked on this package yet, let me know if you'd like me to create a PR.

willemstuursma:
If I am not mistaken, the refund has to be "get" from the Payment. It cannot be retrieved from `v2/refunds/re_xxxxx`. See the `self` URL in the `_links` property. 

I guess for canceling a refund, you would usually have the paymentId on hand right?

Do you have a use case?

sandervanhooft:
Agreed. Retrieving the refund via the payment is fine. I was a bit confused because `$client->refunds` exists (`RefundEndpoint`).

There's however no way (afaik) to retrieve a single refund (by refund-id). Right now you need to manually filter the results from `$payment->refunds()`.

Perhaps add a `$payment->getRefund(...)` helper?

sandervanhooft:
(There's a similar `getMandate('some_id')` helper method on the `Customer` resource.)

Smitsel:
Its allright if you make a PR for it :)

sandervanhooft:
Ok, to be sure, that would require the following to be implemented:

```
RefundEndpoint:
tests/Mollie/API/Endpoints/RefundEndpointTest : testGetForWorks
src/Endpoints/RefundEndpoint : getFor(Payment $payment, $refundId, array $parameters = [])

PaymentResource:
tests/Mollie/API/Resources/Payment : testGetRefundWorks
src/Resources/Payment : getRefund($refundId, array $parameters = [])
```

sandervanhooft:
Will work on this later this week.

Also noted that there's a minor docbloc issue [in the Payment resource](https://github.com/mollie/mollie-api-php/blob/8acb00b90a5a7a2ecfcd11d4608444150d89a220/src/Endpoints/PaymentEndpoint.php#L125), referencing `$filters`, which should be `$data`.

Will take care of this as well.

sandervanhooft:
Hi @Smitsel ,

Working on it...

Shouldn't RefundEndpoint have `payments_refunds` as the resource path?

```
class RefundEndpoint extends EndpointAbstract
{
    protected $resourcePath = "refunds"; // should be "payments_refunds"?
```

```
[...]

PR #242, opened 2018-08-02
Add methods payment->getRefund and refunds->getFor (#235)

sandervanhooft:


tests/Mollie/API/Endpoints/RefundEndpointTest.php
```diff
@@ -91,6 +91,154 @@ public function testCreateRefund()
         $this->assertEquals($documentationLink, $refund->_links->documentation);
     }
 
+    public function testGetRefundForPaymentResource()
+    {
+        $this->mockApiCall(
+            new Request(
+                "GET",
+                "/v2/refunds/re_PsAvxvLsnm"
+                // "/v2/payments/tr_44aKxzEbr8/refunds/re_PsAvxvLsnm"
```
Smitsel: The expectation and actual request Uri should be `/payments/*/refunds/*`

You are right that it should be `payments_refunds` but the problem is that the RefundEndpoint is for the list refunds as shown in: https://docs.mollie.com/reference/v2/refunds-api/list-refunds
Where the result would be GET /refunds.

So what you will need to do is create a PaymentRefundsEndpoint and move the getFor there.

sandervanhooft:
I noticed the naming (singular / plural) is not entirely consistent... What's preferred?

Most is "all singular". So I'll go with `PaymentRefundEndpointTest` and `PaymentRefundEndpoint`.

```bash
CustomerEndpoint
CustomerPaymentsEndpoint # singular Customer, plural Payments
EndpointAbstract
InvoiceEndpoint
MandateEndpoint
MethodEndpoint
PaymentEndpoint
ProfileEndpoint
RefundEndpoint
SettlementsEndpoint # plural
SubscriptionEndpoint


BaseEndpointTest
ChargebackEndpointTest
CustomerEndpointTest
CustomerPaymentEndpointTest # singular
InvoiceEndpointTest
MandateEndpointTest
MethodEndpointTest
PaymentEndpointTest
ProfileEndpointTest
RefundEndpointTest
SettlementEndpointTest # singular
SubscriptionEndpointTest
```

Smitsel:
I would go for Singular.

src/Endpoints/PaymentRefundEndpoint.php
```diff
@@ -0,0 +1,65 @@
...
+     * Retrieves a collection of Refunds from Mollie.
+     *
+     * @param string $from The first refund ID you want to include in your list.
+     * @param int $limit
+     * @param array $parameters
+     *
+     * @return RefundCollection
+     * @throws ApiException
+     */
+    public function page($from = null, $limit = null, array $parameters = [])
```
Smitsel: Should this not be listFor?
Since this should retrieve all refunds for the given payment?

Or maybe remove it and use $payment->refunds() only?

Smitsel: Disregard my first comment, i think it should be removed so we only use `$payment->refunds()` and use the nice _links property.

sandervanhooft: 👍  Removed it.

Smitsel:
Muchos Gracias!

sandervanhooft:
🎉

sandervanhooft:
BTW, what linter / settings are you using? I noticed some differences in this client and the Laravel client vs my local setup.

Smitsel:
Ill send u an email ;)

sandervanhooft:
Thanks!

PR #259, opened 2018-08-30
Add invoiceId property to Settlement

malles:
As per the [get-settlement docs](https://docs.mollie.com/reference/v2/settlements-api/get-settlement) the invoiceId is present on the settlement object. This commit adds the property to the class, so it can be retrieved via this client.

### DIFF
--- a/src/Resources/Settlement.php
+++ b/src/Resources/Settlement.php
@@ -59,7 +59,7 @@ class Settlement extends BaseResource
     /**
      * The ID of the invoice on which this settlement is invoiced, if it has been invoiced.
      *
-     * @var string
+     * @var string|null
      */
     public $invoiceId;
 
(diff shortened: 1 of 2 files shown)


src/Resources/Settlement.php
```diff
@@ -56,6 +56,13 @@ class Settlement extends BaseResource
      */
     public $periods;
 
+    /**
+     * The ID of the invoice on which this settlement is invoiced, if it has been invoiced.
+     *
+     * @var string
```
willemstuursma: The property is nullable, so this should be `string|null`.

tests/Mollie/API/Endpoints/SettlementEndpointTest.php
```diff
@@ -379,6 +381,7 @@ public function testListSettlement()
                             }
                           }
                         },
+                        "invoiceId": "",
```
willemstuursma: If there is no invoice, it will be omitted. It will never be empty string. Can you put a value here or omit it and assert that the invoiceId in the Settlement object is null?

malles:
Thanks for the quick merge!

PR #261, opened 2018-09-03
Added Shipment.update() for tracking info

sandervanhooft:
See https://github.com/mollie/api-documentation/pull/193


Smitsel:
You're fast! Thanks for this!

Smitsel:
@mvdpanne just reminded me on https://github.com/mollie/api-documentation/pull/193
That this needs an example too.

Could you add it and also add it to the api-documentation?

PR #285, opened 2018-10-02
Add new 'authorized' payment status for Klarna payments

Daanvm:
The `authorized` status is currently only possible for _Klarna Pay later_ and _Klarna Slice it_. Payments with these payment methods can only be created with the [Orders API](https://docs.mollie.com/reference/v2/orders-api/create-order). You should [create a Shipment](https://docs.mollie.com/reference/v2/shipments-api/create-shipment) to trigger the capture to receive the money.

### DIFF
--- a/src/Resources/Payment.php
+++ b/src/Resources/Payment.php
@@ -262,6 +262,16 @@ class Payment extends BaseResource
     }
 
     /**
+     * Is this payment authorized?
+     *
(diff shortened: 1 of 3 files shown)


ndijkstra:
@sandervanhooft FYI, we've added a new status for v2 payments. Maybe you need also to change something in the Laravel client?

Issue #284, opened 2018-10-02
Implementation of canceling order lines is outdated


mvdpanne:
## Specifications

  - API Version: v2

## Describe the issue

About a month ago, we have decided to move from a `cancelOrderline` endpoint to a `cancelOrderlines` endpoint. This allows merchants to cancel multiple order lines in one API call, instead of having to do multiple API calls.

There was probably some lack of communication about this, which is our bad. Please see the updated documentation here: https://docs.mollie.com/reference/v2/orders-api/cancel-order-lines. The examples in the docs will need to be updated as well.

@sandervanhooft, could you pick this up?

sandervanhooft:
@mvdpanne No problem. Sure, I'll pick this one up.

@Smitsel 
Do you want to preserve the `$order->cancelLine("odl_dgtxyl", 2);` method as a shorthand for

```php
$order->cancelLines([
    [
        'id' => 'odl_dgtxyl',
        'quantity' => 2,
    ],
]);
```

?

Smitsel:
I think cancelLines is fine.

Maybe we should even take it a step further and take `$data` on the cancel method.
I think that would be most preferred since it is consistent with refund :)

I'll let you decide :)

sandervanhooft:
Tests:
- [ ] Replace `OrderEndpointTest.testCancelOrderLinePartially()` by `.testCancelOrderLines()`
- [ ] Remove `OrderEndpointTest.testCancelOrderLineOnResourceCompletely()`
- [ ] Remove `OrderEndpointTest.testCancelOrderLineOnResourcePartially()`

Source:
- [ ] Remove `OrderLine $orderline->cancel()`
- [ ] Replace `Order $order->cancelLine(...)` by `Order $order->cancelLines($data)`
- [ ] Update `OrderLineEndpoint $orderlines->cancelFor()`

Docs:
- [ ] Update code example `30-cancel-order-line`
- [ ] Update examples at mollie.com reference

sandervanhooft:
Also added `$order->cancelAllLines()`

sandervanhooft:
All done. PRs ready. 😄

PR #286, opened 2018-10-02
Bug fix: Cancel order lines

sandervanhooft:
Solves https://github.com/mollie/mollie-api-php/issues/284 , except for the Mollie docs.


sandervanhooft:
It's not a breaking change but a bug fix actually.

sandervanhooft:
@Smitsel @mvdpanne

PR #300, opened 2018-10-13
Fetch captures for settlement

SanderVanLeeuwen:
The current client is missing a method to fetch captures associated with a settlement. We do need those to keep our administration complete, so I've quickly added this method based on the other methods format.

As a side note; there's a lot of repeating code in this class to fetch different types of resource collections. Is this on purpose to keep it straight forward?


Smitsel:
Thanks @SanderVanLeeuwen

PR #320, opened 2018-11-23
Also check HTTP_X_FORWARDED_PROTO for protocol

stefanvangastel:
In order to facilitate cloud environments like Azure containers where applications run as http servers and are loadbalanced by a httpS reverse proxy an extra check for https was added.


sandervanhooft:
@willemstuursma Can this PR be closed?

Issue #332, opened 2018-12-11
add support for enabling/disabling methods


sweebee:
## Specifications

  - API Version: 2

## Describe the issue

its just integrated in the API, would be nice to use in this wrapper.

https://docs.mollie.com/reference/v2/profiles-api/enable-method

willemstuursma:
@sandervanhooft could you give this a shot?

If not possible, the exception should have a method that can easily give you the URL where the merchant can enable the payment method.

sandervanhooft:
Sure thing @willemstuursma !

sandervanhooft:
Shouldn't the profile method response include a `_link` to the `profile` as well?

willemstuursma:
@MollieRick could you look into this?

MollieRick:
@sandervanhooft @willemstuursma I will add it.

sandervanhooft:
> @sandervanhooft could you give this a shot?
> 
> If not possible, the exception should have a method that can easily give you the URL where the merchant can enable the payment method.

```
HTTP/1.1 422 Unprocessable Entity
Content-Type: application/hal+json; charset=utf-8

{
    "status": 422,
    "title": "Unprocessable Entity",
    "detail": "Can not enable Credit card via the API. Please go to the dashboard to enable this payment method.",
    "_links": {
         "dashboard": {
                "href": "https://www.mollie.com/dashboard/settings/profiles/pfl_v9hTwCvYqw/payment-methods",
                "type": "text/html"
         },
         "documentation": {
                "href": ":doc:`https://docs.mollie.com/guides/handling-errors </guides/handling-errors>`",
                "type": "text/html"
         }
    }
}
```

The API exception message is passed through by default. You want me to catch this and amend the message to include the link?

I think a better solution may be to amend the message at the API. You'll probably need it in all integrations.

Instead of:
```
Can not enable Credit card via the API. Please go to the dashboard to enable this payment method.
```

Do something like:
```
Can not enable Credit card via the API. Please enable it here: https://www.mollie.com/dashboard/settings/profiles/pfl_v9hTwCvYqw/payment-methods
```

willemstuursma:
I would like it to be possible for integrators to easily create a clickable link if enabling automatically fails.

So the exception should have some `getDashboardLink()` method.

sandervanhooft:
Interesting idea @willemstuursma , I'm on it!

sandervanhooft:
[...]

PR #334, opened 2018-12-12
Profile methods

sandervanhooft:
Solves #332.

Adding:
- `$profile->enableMethod($methodId, $data) : Method`
- `$profile->disableMethod($methodId, $data) : null`

### DIFF
--- a/tests/Mollie/API/Endpoints/ProfileMethodEndpointTest.php
+++ b/tests/Mollie/API/Endpoints/ProfileMethodEndpointTest.php
@@ -39,10 +39,6 @@ class ProfileMethodEndpointTest extends BaseEndpointTest
                         "documentation": {
                             "href": "https://docs.mollie.com/reference/v2/profiles-api/activate-method",
                             "type": "text/html"
-                        },
-                        "profile": {
-                             "href": "https://api.mollie.com/v2/profiles/pfl_v9hTwCvYqw",
-                             "type": "application/hal+json"
                         }
                     }
                 }'
@@ -69,12 +65,6 @@ class ProfileMethodEndpointTest extends BaseEndpointTest
             "https://docs.mollie.com/reference/v2/profiles-api/activate-method",
             "text/html",
             $method->_links->documentation
-        );
-
-        $this->assertLinkObject(
-            "https://api.mollie.com/v2/profiles/pfl_v9hTwCvYqw",
-            "application/hal+json",
-            $method->_links->profile
         );
 
     }
--- a/tests/Mollie/API/Endpoints/ProfileMethodEndpointTest.php
+++ b/tests/Mollie/API/Endpoints/ProfileMethodEndpointTest.php
@@ -27,18 +27,22 @@ class ProfileMethodEndpointTest extends BaseEndpointTest
                     "id": "bancontact",
                     "description": "Bancontact",
                     "image": {
-                        "size1x": "https://mollie.dev/external/icons/payment-methods/bancontact.png",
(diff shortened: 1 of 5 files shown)


tests/Mollie/API/Endpoints/ProfileMethodEndpointTest.php
```diff
@@ -0,0 +1,137 @@
...
+            ),
+            new Response(
+                201,
+                [],
+                '{
+                    "resource": "method",
+                    "id": "bancontact",
+                    "description": "Bancontact",
+                    "image": {
+                        "size1x": "https://mollie.dev/external/icons/payment-methods/bancontact.png",
```
MollieRick: This is an error in the docs what we're fixing now. Please use www.mollie.com :-)

tests/Mollie/API/Endpoints/ProfileMethodEndpointTest.php
```diff
@@ -0,0 +1,137 @@
...
+                        "size2x": "https://mollie.dev/external/icons/payment-methods/bancontact%402x.png",
+                        "svg": "https://mollie.dev/external/icons/payment-methods/bancontact.svg"
+                    },
+                    "_links": {
```
MollieRick: We're adding `profile` to the link-set.

```
"profile": {
     "href": "https://api.mollie.com/v2/profiles/pfl_xxxx",
     "type": "application/hal+json"
}
```

MollieRick:
@sandervanhooft Because this endpoint has a lot of impact on the merchant, we expect that developers also mention the prices when they use this endpoint. Can you also add the `?include=pricing` on the Methods API? :)

sandervanhooft:
> @sandervanhooft Because this endpoint has a lot of impact on the merchant, we expect that developers also mention the prices when they use this endpoint. Can you also add the `?include=pricing` on the Methods API? :)

@MollieRick Noted! I'm working on the dev-friendly exception handling first.

tests/Mollie/API/Endpoints/ProfileMethodEndpointTest.php
sandervanhooft: Done!

tests/Mollie/API/Endpoints/ProfileMethodEndpointTest.php
sandervanhooft: Done, thanks!

sandervanhooft:
> @sandervanhooft Because this endpoint has a lot of impact on the merchant, we expect that developers also mention the prices when they use this endpoint. Can you also add the `?include=pricing` on the Methods API? :)

@MollieRick Can you provide me with an example response?

MollieRick:
> @MollieRick Can you provide me with an example response?

Sure. You'll find an example here in the [GET Methods doc](https://docs.mollie.com/reference/v2/methods-api/get-method).

A method can contain multiple pricing rates.

sandervanhooft:
> > @MollieRick Can you provide me with an example response?
> 
> Sure. You'll find an example here in the [GET Methods doc](https://docs.mollie.com/reference/v2/methods-api/get-method).
> 
> A method can contain multiple pricing rates.

Ah ok, this is a different API, overlooked that. Will work on it in a separate PR/issue.

sandervanhooft:
@MollieRick I think this PR is ready now. Will handle ApiException separately.

Smitsel:
Good call to do the Exception handling in a different PR. It's better if @MollieRick takes a look at this one when he is back from vacation next week.

MollieRick:
LGTM!

sandervanhooft:
Thanks @Smitsel and @MollieRick ! 😄

PR #342, opened 2019-01-03
Profiles.getCurrent()

sandervanhooft:


sandervanhooft:
@Smitsel Do we want such a method, if so, should it be called `getCurrent` or `getMe`?

tests/Mollie/API/Endpoints/ProfileEndpointTest.php
```diff
@@ -99,6 +99,95 @@ public function testGetProfile()
...
+                        "self": {
+                            "href": "https://api.mollie.com/v2/profiles/pfl_ahe8z8OPut",
+                            "type": "application/hal+json"
+                        },
+                        "chargebacks": {
+                            "href": "https://api.mollie.com/v2/chargebacks?profileId=pfl_ahe8z8OPut",
+                            "type": "application/hal+json"
+                        },
+                        "methods": {
+                            "href": "https://api.mollie.com/v2/methods?profileId=pfl_ahe8z8OPut",
```
Smitsel: I do not have the `?profileId=` query string parameter in my actual response when i call `/v2/profiles/me`

src/Endpoints/ProfileEndpoint.php
```diff
@@ -50,19 +50,32 @@ public function create(array $data = [], array $filters = [])
...
 
+    /**
+     * Retrieve the current Profile from Mollie.
+     *
+     * @param array $parameters
+     *
+     * @return \Mollie\Api\Resources\BaseResource|\Mollie\Api\Resources\Profile
+     * @throws ApiException
+     */
+    public function getCurrent(array $parameters = [])
```
Smitsel: I think this name is fine ! :)

tests/Mollie/API/Endpoints/ProfileEndpointTest.php
sandervanhooft: I see, the [docs](https://docs.mollie.com/reference/v2/profiles-api/get-profile-me#example) agree with you!

So I should adhere to the docs, or have we discovered an API issue here? (Don't think so)

sandervanhooft: I'll remove the link params for now, let me know if things change :)

sandervanhooft:
@Smitsel Fixed

PR #343, opened 2019-01-03
Added missing organization attributes locale and email.

vernondegoede:


sandervanhooft:
Thanks @vernondegoede,

Did you check if it matches the test responses?

vernondegoede:
Thanks, missed that. I've updated the mocked response and added an assertion.

PR #346, opened 2019-01-08
Adding method for enable and disable payment method on profile me

MollieRick:


sandervanhooft:
Not sure what this solves? 🤔

Smitsel:
Hey @sandervanhooft, yesterday this was implemented on our side and we encountered an error.

What we did was use the `profiles/me` endpoint to get the used profile and then tried enabling a method on that with the `enableMethod` on the Profile resource.

However, that uses the profile token to do the enable call like this:
`POST v2/profiles/pfl_xxxx/methods/ideal`
But this is an oAuth only endpoint and thus we got some errors. 

We should use the profiles/me variation of the method enabling endpoint:
`POST v2/profiles/me/methods/ideal`

Can you take over this PR and add the profiles/me enable endpoint to the client? With some tests and stuff?

sandervanhooft:
Ah now I see! Thanks @Smitsel for clearing this up. Plotted it in this matrix below to fully understand what's going on.

|Action|Method|URL|Authentication|
|-|-|-|-|
| Retrieve a Profile | GET | `v2/profiles/*id*` | Organization access tokens, App access tokens |
| Retrieve current Profile | GET | `v2/profiles/me` | API keys |
| Enable payment method | POST | `v2/profiles/*id*/methods/*method*` | Organization access tokens, App access tokens |
| Enable payment method on current Profile | POST | `v2/profiles/me/methods/*method*` | API keys |

I think it's best to somehow add a CurrentProfile Resource, extending the Profile and override specific `me` behaviour there.

I'll look into it.

Smitsel:
Sounds good @sandervanhooft! Thanks for picking this up.

sandervanhooft:
@MollieRick was working on the right direction with this PR, but I can't find a way to overtake it.

Moving this to a new PR.

PR #347, opened 2019-01-09
Current profile methods

sandervanhooft:


sandervanhooft:
Continues work from #346 .

sandervanhooft:
I think this covers the requirements.

Also made sure that the CurrentProfile resource is returned when `$mollie->profiles->get('me')` is called.

Issue #335, opened 2018-12-13
Add ?include=pricing to Methods API


sandervanhooft:
## Specifications

  - API Version: 2

## Describe the issue

As suggested by @MollieRick in #332, we should include support for `?include=pricing` option in the Methods API. I'll come up with a PR later.

sandervanhooft:
[Example response](https://docs.mollie.com/reference/v2/methods-api/get-method) (includes `issuers` as well):

```
HTTP/1.1 200 OK
Content-Type: application/hal+json

{
     "resource": "method",
     "id": "ideal",
     "description": "iDEAL",
     "image": {
         "size1x": "https://www.mollie.com/external/icons/payment-methods/ideal.png",
         "size2x": "https://www.mollie.com/external/icons/payment-methods/ideal%402x.png",
         "svg": "https://www.mollie.com/external/icons/payment-methods/ideal.svg"
     },
     "issuers": [
         {
             "resource": "issuer",
             "id": "ideal_ABNANL2A",
             "name": "ABN AMRO",
             "image": {
                 "size1x": "https://www.mollie.com/external/icons/ideal-issuers/ABNANL2A.png",
                 "size2x": "https://www.mollie.com/external/icons/ideal-issuers/ABNANL2A%402x.png",
                 "svg": "https://www.mollie.com/external/icons/ideal-issuers/ABNANL2A.svg"
             }
         },
         {
             "resource": "issuer",
             "id": "ideal_ASNBNL21",
             "name": "ASN Bank",
             "image": {
                 "size1x": "https://www.mollie.com/external/icons/ideal-issuers/ASNBNL21.png",
                 "size2x": "https://www.mollie.com/external/icons/ideal-issuers/ASNBNL21%402x.png",
                 "svg": "https://www.mollie.com/external/icons/ideal-issuers/ASNBNL21.svg"
             }
         },
         { },
         { }
     ],
     "pricing": [
         {
             "description": "The Netherlands",
             "fixed": {
                 "value": "0.29",
                 "currency": "EUR"
             },
             "variable": "0"
         }
     ],
     "_links": {
         "self": {
             "href": "https://api.mollie.com/v2/methods/ideal",
             "type": "application/hal+json"
         },
         "documentation": {
             "href": "https://docs.mollie.com/reference/v2/methods-api/get-method",
             "type": "text/html"
         }
     }
 }
```

MollieRick:
Small note: `?include=pricing` will **not** include issuers by default.

[...]

PR #348, opened 2019-01-10
Pricing resource

sandervanhooft:
#335 

Refactored other nested resource accessors to use the newly introduced `createBaseResourceCollection` and `createCursorResourceCollection` helpers.

### DIFF
--- a/src/Resources/BaseResource.php
+++ b/src/Resources/BaseResource.php
@@ -18,4 +18,27 @@ abstract class BaseResource
     {
         $this->client = $client;
     }
+
+    /**
+     * Create a resource collection from an array.
+     *
+     * @param array $input
+     * @param string $resourceClass The full class namespace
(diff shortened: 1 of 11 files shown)


src/Resources/BaseResource.php
```diff
@@ -18,4 +18,50 @@ public function __construct(MollieApiClient $client)
     {
         $this->client = $client;
     }
+
```
MollieRick: Maybe is it, indeed, better to move this to the `ResourceFactory` to keep these functions together.

MollieRick: Maybe is it, indeed, better to move this to the ResourceFactory to keep these functions together.

sandervanhooft: Agreed. Will refactor later today.

PR #353, opened 2019-01-17
Order payments() helper

sandervanhooft:
Retrieves the nested payments as a PaymentsCollection.

Requires the Order to be retrieved using the `include=payments` flag.


sandervanhooft:
Just noticed that according the docs `include` flag should be `embed` in the Order endpoint case. Seems inconsistent with Payment endpoint, so followed up on Mollie Slack. Stay tuned.

PR #358, opened 2019-02-03
Direct calls for subresources

sandervanhooft:


willemstuursma:
@Smitsel has some time off this week @sandervanhooft, I'll ask @ndijkstra to look into it.

willemstuursma:
Guess this warrants some API docs updates too, @sandervanhooft.

sandervanhooft:
@willemstuursma Will update the docs later this week, thanks!

sandervanhooft:
@willemstuursma 

Started working on the docs, but I'm not sure this is such a great idea (not in every case anyway). Direct calls feel more like "advanced-level stuff", cluttering up the more readable indirect calls.

In this case, it's still sort of readable:

```php
      $mollie = new \Mollie\Api\MollieApiClient();
      $mollie->setApiKey("test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM");

      // on the payment
      $payment = $mollie->payments->get("tr_WDqYK6vllg");
      $chargeback = $payment->getChargeback("chb_n9z0tp");

      // or directly
      $chargeback = $mollie->chargebacks->getForId("tr_WDqYK6vllg", "chb_n9z0tp");
```

But in other cases, it clutters up too much IMHO. I.e. this one:
```php
$mollie = new \Mollie\Api\MollieApiClient();
$mollie->setApiKey("test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM");

// List chargebacks for a single payment
$payment = $mollie->payments->get("tr_7UhSN1zuXS");
$chargebacks = $payment->chargebacks();

// or directly
$chargebacks = $mollie->paymentChargebacks->listForId("tr_7UhSN1zuXS");

// List chargebacks across all payments on the payment profile
// (For all chargebacks on the organizations, use an OAuth or Organization access token.)
$all_chargebacks = $mollie->chargebacks->page();
```

`listForId()` does not exist yet btw, currently `$payment->chargebacks()` uses a direct http call.

willemstuursma:
`listForId` should be `listForPaymentId`, does that help?

sandervanhooft:
About `getForId()`:

I noticed that in pre-existing subresource endpoints `getForId` was being used, instead of including the parent resource name (`getForPaymentId`), so decided to stick with that.

A reference to the parent resource is already included in the endpoint accessor (`paymentChargebacks`), _implying_ what the `getForId` call is about.

I'm ok with adding `getForPaymentId` or `getForParentId` if that clears it up; it would then be like this:

```php
$chargebacks = $mollie->paymentChargebacks->listForId("tr_7UhSN1zuXS");

// vs

$chargebacks = $mollie->paymentChargebacks->listForPaymentId("tr_7UhSN1zuXS");

// vs

$chargebacks = $mollie->paymentChargebacks->listForParentId("tr_7UhSN1zuXS");
```

As these calls `getForId` calls are already released (some with the initial v2 release it seems), these new methods would be added as aliases.

Smitsel:
I think just keeping the `getForId` is fine.

It might be confusing when the endpoint accessor is named `paymentChargebakcs`. The ID could be about the payment or the chargeback.

But if we make sure that the parameter name in the signature of `getForId` is $paymentId it will be clear enough for integrators.

Smitsel:
Also, i believe we should not document the direct calls as examples.

We want to encourage our integrators to use the client from the parent resource object.
But that's just my 2 cents.

sandervanhooft:
Just discussed with Martijn: we'll leave the docs as-is for readability. Advanced users will be able to find the direct call endpoint methods. Thanks @willemstuursma and @Smitsel !

PR #361, opened 2019-02-14
Travis: allow PHP nightly to fail and switch to xenial images

firstred:
Removing PHP nightly from the original array, but keeping it under `allow_failures` will allow it to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,38 +2,37 @@ language: php
 
 dist: trusty
 
-php:
-- 5.6
-- 7.0
-- 7.1
-- 7.2
-- 7.3
-
 matrix:
   fast_finish: true
   allow_failures:
-  - php: nightly
+    - php: nightly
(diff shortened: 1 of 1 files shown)


firstred:
Nope, that does not work. We should go for matrix builds. New commits coming soon!

firstred:
Using a build matrix for all PHP versions should fix it.

firstred:
Hmmz, these are the same results I see in my repo. It turns out that testing on xenial is more stable, which makes sense since trusty has been deprecated for a while already.

One more test, just to see if everything can be moved to the xenial images.

firstred:
Alrighty, that works! Enjoy a new, more stable testing experience! :slightly_smiling_face:

willemstuursma:
Cool, opened https://github.com/composer/ca-bundle/pull/34.

PR #363, opened 2019-02-16
Improved first payment method mapper

sandervanhooft:
Should return `null` for non-first-payment methods.

This requires introducing the `FirstPaymentMethod` type, along with some methods (`all`, `exists`).

### DIFF
--- a/src/Types/FirstPaymentMethod.php
+++ b/src/Types/FirstPaymentMethod.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Mollie\Api\Types;
+
+class FirstPaymentMethod
+{
+    const CREDITCARD = 'creditcard';
+    const BANCONTACT = 'bancontact';
+    const BELFIUS = 'belfius';
+    const EPS = 'eps';
+    const GIROPAY = 'giropay';
+    const IDEAL = 'ideal';
+    const INGHOMEPAY = 'inghomepay';
+    const KBC = 'kbc';
+    const SOFORT = 'sofort';
+
+    /**
+     * Retrieve all methods that can be used for a first payment.
+     *
+     * @return array
+     */
+    public static function all()
+    {
+        return [
+            static::CREDITCARD,
+            static::BANCONTACT,
+            static::BELFIUS,
+            static::EPS,
+            static::GIROPAY,
+            static::IDEAL,
+            static::INGHOMEPAY,
+            static::KBC,
+            static::SOFORT,
+        ];
+    }
+
+    /**
+     * Whether the method can be used as a first payment method.
+     *
+     * @param $method
+     * @return bool
+     */
+    public static function exists($method)
+    {
+        return in_array($method, static::all());
+    }
+}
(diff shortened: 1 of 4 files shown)


tests/Mollie/API/Types/FirstPaymentMethodTest.php
```diff
@@ -0,0 +1,32 @@
...
+namespace Tests\Mollie\API\Types;
+
+use Mollie\Api\Resources\Payment;
+use Mollie\Api\Types\FirstPaymentMethod;
+use Mollie\Api\Types\PaymentMethod;
+use PHPUnit\Framework\TestCase;
+
+class FirstPaymentMethodTest extends TestCase
+{
+    public function testGetAll()
```
willemstuursma: This isn't worthy of a test. Its a so called change detector test https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html

src/Types/FirstPaymentMethod.php
```diff
@@ -0,0 +1,47 @@
...
+        ];
+    }
+
+    /**
+     * Whether the method can be used as a first payment method.
+     *
+     * @param $method
+     * @return bool
+     */
+    public static function exists($method)
```
willemstuursma: It would be simpler if this method were to be moved to the existing Method object. Saves a class and all the constants.

tests/Mollie/API/Types/FirstPaymentMethodTest.php
sandervanhooft: Agreed. Wrote the test before knowing what the implementation would be. Will clear it up. Thanks for the nice read 😄.

sandervanhooft: BTW, I used to test this with `assertArraySubset` (which would have made it less of a change detector test), but that assertion method is marked as deprecated now. Any suggestion?

src/Types/FirstPaymentMethod.php
sandervanhooft: @willemstuursma As `isFirstPaymentMethod($method)` you mean?

Would require something like a static `firstPaymentMethods()` as well.

willemstuursma:
I'm not sure about this. If we add another payment method as first (PayPal obviously comes to mind) users of this class won't be able to use it automatically. What are you trying to accomplish?

sandervanhooft:
@willemstuursma I _had_ a use case for this (Cashier), but I'm refactoring so much mandate functionality right now, I'm not sure it's required when I'm done.

The base functionality for mapping first payment methods to resulting mandate methods is already released (2.6 I believe), and should be improved as it is.

Will get back to you once I'm done refactoring.

sandervanhooft:
The use case is for multi-step recurring payments, without persisting the `mandateId` in the client for future usage:

1. Perform a first payment, with a `method` defined, i.e. `ideal`.
2. Perform a mandated payment, _without_ providing the `mandateId`, but _with_ the corresponding `method` provided in the payload. In the case of `ideal`, this `method` would be `directdebit`. The method mapper is very helpful for this.

The approach above allows the client to select a specific method without tying it to a specific mandate.

In situ Cashier, I've moved away from this approach. I'm now persisting the `mandateId` for future usage. This also allows features like `swapMandate()` to be implemented later on.

Issue #364, opened 2019-02-26
Methods: LIST vs ALL


sandervanhooft:
## Specifications

  - API Version: v2

## Describe the issue

The Mollie API v2 offers these two endpoints:

- `LIST /v2/methods`: Retrieve all active payment methods. The results are not paginated. In test mode, this includes `pending` methods.
- `ALL /v2/methods/all`: Retrieve all payment methods offered by Mollie, for which the requesting organization is eligible. This includes both activated and not (yet) activated payment methods.

For this client, we're facing two issues relating to these endpoints:

1. Currently, this client has not implemented a call to the ALL methods endpoint.
2. More unfortunately `$mollie->methods->all()` is already in use for calling the LIST endpoint.

Based on the discussion in the Mollie Slack channel, I'll outline a solution and provide a matching PR.

sandervanhooft:
* [ ] Mark `$mollie->methods->all()` as `deprecated`
* [ ] For `LIST /v2/methods` provide `$mollie->methods->allActive()`
* [ ] For `ALL /v2/methods` provide `$mollie->methods->allAvailable()`
* [ ] Improve docs:
    * [ ] LIST methods page ([link](https://docs.mollie.com/reference/v2/methods-api/list-methods))
    * [ ] ALL methods page ([link](https://docs.mollie.com/reference/v2/methods-api/list-all-methods))

PR #365, opened 2019-02-26
Methods: all, allAvailable, allActive

sandervanhooft:
See Issue #364

* [x] Mark `$mollie->methods->all()` as `deprecated`
* [x] For `LIST /v2/methods` provide `$mollie->methods->allActive()`
* [x] For `ALL /v2/methods/all` provide `$mollie->methods->allAvailable()`


sandervanhooft:
@Smitsel final check: [the docs](https://docs.mollie.com/reference/v2/methods-api/list-methods) mention both `active` and `enabled`.

Should the method be named `allEnabled()` instead of `allActive()`?

PR #368, opened 2019-03-05
Filter mandates by status, housekeeping

sandervanhooft:
Adds a `whereStatus($status)` filter method to `MandateCollection`.

If desired, I can also add shortcuts for `whereStatusValid()` / `whereStatusInvalid()` / `whereStatusPending()`.

Also, `whereStatus($status)` could be an interesting candidate for a trait, so it also can be used with other resources (i.e. `Payment`).

@Smitsel What do you think?

PS. This would be the first trait in the package, not sure if there's any Mollie coding style guide against that. The package is PHP 5.6 and up, so no problem there.

### DIFF
--- a/src/Endpoints/EndpointAbstract.php
+++ b/src/Endpoints/EndpointAbstract.php
@@ -141,7 +141,7 @@ abstract class EndpointAbstract
      * @return BaseCollection
      * @throws ApiException
      */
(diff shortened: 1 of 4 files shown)


sandervanhooft:
@mollierick What do you think of my comments above?

PR #382, opened 2019-04-18
Add onboarding endpoint

LauLaman:
Add implementation for the new [onboarding endpoint](https://docs.mollie.com/reference/v2/onboarding-api/get-onboarding-status)

_Remark:_
The onboarding endpoint has no collection of resources therefore i extracted that logic to  `CollectionEndpointAbstract`

### DIFF
--- a/src/Endpoints/ChargebackEndpoint.php
+++ b/src/Endpoints/ChargebackEndpoint.php
@@ -6,7 +6,7 @@ use Mollie\Api\Exceptions\ApiException;
 use Mollie\Api\Resources\Chargeback;
 use Mollie\Api\Resources\ChargebackCollection;
 
-class ChargebackEndpoint extends EndpointAbstract
+class ChargebackEndpoint extends CollectionEndpointAbstract
 {
     protected $resourcePath = "chargebacks";
 
--- a/src/Endpoints/CollectionEndpointAbstract.php
+++ b/src/Endpoints/CollectionEndpointAbstract.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Mollie\Api\Endpoints;
+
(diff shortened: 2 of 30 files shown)


sandervanhooft:
@LauLaman 

Thanks!

Ok, just to get things clear. There are two things going on here:

1. Adding the Onboarding Endpoint.
2. Splitting the former EndpointAbstract into a CollectionEndpointAbstract and a stripped-down EndpointAbstract, so the latter better fits the new Onboarding Endpoint.

Correct?

LauLaman:
@sandervanhooft 

1. Yes, one remark:
Unfortunately i had to overwrite the `rest_read` and `rest_create` methods since the onboarding Api does not play nice with the current implementation 

2. Yeah i did not wanna implement the abstract method `getResourceCollectionObject` in the Onboarding endpoint that does do nothing.  So i moved that to its own abstract class

sandervanhooft:
Ok clear @LauLaman , thanks. I agree your refactor provides more clarity.

I'll look into this; there are some other places where `rest_read` and `rest_create` are circumvented.
I.e. for calling `/methods/all`.

ChristianVermeulen:
@sandervanhooft In order to be able to actually use this I've added the permission constants in the oauth repo: https://github.com/mollie/oauth2-mollie-php/pull/15.

However, the last time tests ran for that repo was 6 months ago and it now seems like the php nightly tests are failing due to changes at travis. There is actually a second PR open with even more permission constants.

Perhaps you guys can add this with suitable priority on the board to get this merged as well? :D

sandervanhooft:
Hi @ChristianVermeulen,

Thanks for getting me in the loop, wasn't involved yet in the oauth package. Will look into this this afternoon!

PR #387, opened 2019-05-14
 Apple Pay support

sandervanhooft:


sandervanhooft:
@mollierick I think this is all impact there is(?)

sandervanhooft:
> I think this is all impact there is(?)

> I think as well!

`This.simplicity === Mollie Powerrrrr`

PR #391, opened 2019-06-06
Add phpstan to CI

nederdirk:
As suggested by @willemstuursma, here is a patch that adds phpstan to the travis definition. To pass phpstan level 1, I also had to make a couple of changes, mainly in type definitions and erroneous function calls.

### DIFF
--- a/.phpstan.ignoreErrors.neon
+++ b/.phpstan.ignoreErrors.neon
@@ -1,0 +1,14 @@
+parameters:
+    ignoreErrors:
+        -
+            message: '/Undefined variable: \$mollie/'
+            paths:
+                - %currentWorkingDirectory%/examples
+        -
+            message: '/Function database_(write|read) not found/'
+            paths:
+                - %currentWorkingDirectory%/examples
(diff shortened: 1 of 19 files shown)


sandervanhooft:
Nice! Can we help contributors run this locally too? (I.e. instructions, pre-commit hook) I for one haven't used it before.

sandervanhooft:
Some issues with PHP versions 5.6, 7 and 7.1. @nederdirk is this something that you can fix?

I'd love to keep full backwards compatibility.

szepeviktor:
PHPStan should be ran only once, for example with PHP 7.3
The code it checks could run on any PHP version.

nederdirk:
@sandervanhooft I finally have gotten around to getting this done. It took some awkward fiddling with `composer.json` and `.travis.yml` to get it to work with old PHP versions

sandervanhooft:
@nederdirk That's great, thanks!

szepeviktor:
Thank you.

PR #401, opened 2019-08-08
Fix minor mistakes in endpoint tests

Pimm:
Hi everyone, you look great! Did you do something with your hair?

As I was porting tests to the Node.js-client, I noticed some things that I think are mistakes.

**Please verify**.

### DIFF
--- a/tests/Mollie/API/Endpoints/CustomerPaymentEndpointTest.php
+++ b/tests/Mollie/API/Endpoints/CustomerPaymentEndpointTest.php
@@ -26,8 +26,8 @@ class CustomerPaymentEndpointTest extends BaseEndpointTest
                       "currency":"EUR"
                     },
                     "description": "My first API payment",
-                    "redirectUrl": "https://example.org/redirect",
-                    "webhookUrl": "https://example.org/webhook",
+                    "redirectUrl": "http://example.org/examples/payment/03-return-page.php?order_id=1234",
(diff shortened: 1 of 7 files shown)


sandervanhooft:
I prefer the current `webhookUrl` and `redirectUrl` for readability. Can you revert these?

The other changes look good to me!

Pimm:
> I prefer the current `webhookUrl` and `redirectUrl` for readability. Can you revert these?

I personally don't care what the URLs are. It's just that currently, the URLs which are `POST`ed are not equal to the ones which are replied.

(Check out [CustomerPaymentEndpointTest.php](https://github.com/mollie/mollie-api-php/blob/master/tests/Mollie/API/Endpoints/CustomerPaymentEndpointTest.php#L16): it `POST`s `https://example.org/webhook` and then replies `http://example.org/examples/payment/02-webhook-verification.php`).

vernondegoede:
I agree with @sandervanhooft – I'd prefer the old URLs. While you're at it, could you update the URLs in the assertions as well?

Pimm:
Sure. Again, I am not advocating for one URL or the other. What I wanted to fix was the test `POST`ing `https://example.org/webhook` and then replying `http://example.org/examples/payment/02-webhook-verification.php`. Those two should be equal, whatever they might be.

sandervanhooft:
LGTM, thanks @Pimm !

PR #405, opened 2019-09-01
Update payment Bug

BartFijneman:
When using the mollie api php library you can't update a payment.

According to the docs the following code should update a payment, but it throws an error which states: 
'Uncaught Error: Call to undefined method Mollie\Api\Resources\Payment::update()'

`$payment = $mollie->payments->get("tr_7UhSN1zuXS");

$payment->description = "Order #98765";
$payment->redirectUrl = "https://example.org/webshop/order/98765/";
$payment->webhookUrl = "https://example.org/webshop/payments/webhook/";
$payment->metadata = ["order_id" => "98765"];

$payment = $payment->update();`


sandervanhooft:
Fix was already prepared, sorry. (And it's not a bug, just a not yet implemented feature 😎 )

BartFijneman:
Thanks for adding the feature. There's only one problem. If you want to update the payment when its paid it will throw an error because the redirecturl can't be changed. You need to add a check to see if the payment is paid before adding redirecturl to the body.

sandervanhooft:
Ah I think I get your point now. You don't intend to modify the `redirectUrl`, but it's sent anyway, causing an issue.

If for example you update the `webhookUrl` field after the payment has been paid, the client will still send the `redirectUrl` to the API causing an error.

I'll check with the team if we should deal with this in the client(s) or the API.

sandervanhooft:
@BartFijneman the issue has been fixed in the Mollie API.

Issue #296, opened 2018-10-10
Feature request: Allow to determine charged payment fee for a specific payment


holtkamp:
It would be nice to be able to determine the payment fee / transaction cost that was charged for a _specific_ Payment X.

It currently IS possible to acquire the total costs of processed Payments using a certain method, but for transaction costs which involve a percentual part, it would be nice to be able to determine the exact charges of a _particular_ Payment X.

**Observations**
When using the various API endpoints:

- The Payment X object does not contain this information
- The Settlement Y object that "covers" Payment X only contains the cumulative costs of all Payments in the Settlement, for example:
``` 
["costs"] => array(2) {
        [0] => object(stdClass)#23225 (5) {
          ["description"] => string(5) "iDEAL"
          ["method"] => string(5) "ideal"
          ["count"] => int(44)
          ["rate"] => object(stdClass)#23226 (2) {
            ["fixed"] => string(6) "0.2900"
            ["percentage"] => NULL
          }
          ["amount"] => object(stdClass)#23227 (3) {
            ["net"] => string(7) "12.7600"
            ["vat"] => string(6) "2.6796"
            ["gross"] => string(7) "15.4396"
          }
        }
        [1] => object(stdClass)#23228 (5) {
          ["description"] => string(6) "PayPal"
          ["method"] => string(6) "paypal"
          ["count"] => int(1)
          ["rate"] => object(stdClass)#23229 (2) {
            ["fixed"] => string(6) "0.1000"
            ["percentage"] => NULL
          }
          ["amount"] => object(stdClass)#23230 (3) {
            ["net"] => string(6) "0.1000"
            ["vat"] => string(6) "0.0210"
            ["gross"] => string(6) "0.1210"
          }
        }
      }

```
- The Invoice Z object that "covers" Settlement Y (also) only contains the cumulative costs of all involved Payments in the same format as the snippet above.

Especially for payments that involve percentual charges, the *actually* charged costs are required for administration and invoicing purposes. This particularly applies for payment methods with "different" percentages, like the MasterCard and Visa CreditCards:

<img width="940" alt="screenshot 2018-10-10 at 14 51 22" src="https://user-images.githubusercontent.com/776405/46737545-125f4c80-cc9c-11e8-81b4-67a04e29a3e8.png">

*Disclaimer*
Maybe the information can already be extracted from the API, could not find it though...

holtkamp:
[...]

PR #409, opened 2019-09-23
first implementation of getCosts()

christopheg:
Calculate costs based on MethodPrice
rounding is done on 3 decimals
Currently, only the following methods are supported:
- creditcard/american-express
- creditcard/intra-eu
- creditcard/business
- bancontact
- inghomepay
- kbc
- belfius
- banktransfer
- ideal

Only available if transaction has status 'paid'

Refund costs are not available since they cannot be retrieved via any other API call

To use it, simply do:

> 	$payment = $mollie->payments->get($transaction_id);
> 	$payment_costs = $payment->getCosts()->payment;
> 	$refund_costs = $payment->getCosts()->refund; // Not yet supported

### DIFF
--- a/src/Resources/MethodPriceCollection.php
+++ b/src/Resources/MethodPriceCollection.php
@@ -11,4 +11,21 @@ class MethodPriceCollection extends BaseCollection
     {
         return null;
     }
+
+    /**
+     * Get a specific MethodPrice.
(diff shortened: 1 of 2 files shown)


sandervanhooft:
Thanks for the effort @christopheg .

Closing this PR for now, given the discussion in issue #296. Let me know if this needs to be reopened.

PR #416, opened 2019-10-16
Add Mollie_API_Object_Payment::$redirectUrl

holtkamp:
As described in https://docs.mollie.com/reference/v1/payments-api/create-payment#ideal

When _loading_ Payments from the API, the "redirectUrl" is part of the `Payment::$links` property. However when _creating_ a Payment, this property **is** used...

### DIFF
--- a/src/Mollie/API/Object/Payment.php
+++ b/src/Mollie/API/Object/Payment.php
@@ -223,6 +223,13 @@ class Mollie_API_Object_Payment
 	public $mandateId;
 
 	/**
+	 * The URL the customer will be redirected to after the payment process.
+	 *
+	 * @var string|null
+	 */
+	public $redirectUrl;
+	
+	/**
 	 * The identifier referring to the settlement this payment was settled with.
 	 *
 	 * @example stl_BkEjN2eBb


sandervanhooft:
Hi @holtkamp,

Can you help me understand this PR?

The payment is created using an array, not the resource object.

I dived a bit into the `v1` API using the iDeal issuer as you are referring to, but even then this (curl) create payment request:

```bash
curl -X POST https://api.mollie.com/v1/payments \
    -H "Authorization: Bearer test_dHar4XY7LxsDOtmnkVtjNVWXLSlXsM" \
    -d "amount=10.00" \
    -d "description=My first payment" \
    -d "issuer=ideal_INGBNL2A" \
    -d "redirectUrl=https://webshop.example.org/order/12345/" \
    -d "webhookUrl=https://webshop.example.org/payments/webhook/" \
    -d "metadata[order_id]=12345"
```

results in no `redirectUrl` attribute on the Payment (except for the `links`):

```json
{  
   "resource":"payment",
   "id":"tr_wbyGywdAPb",
   "mode":"test",
   "createdDatetime":"2019-10-17T17:53:43.0Z",
   "status":"open",
   "canBeCancelled":false,
   "expiryPeriod":"PT15M",
   "amount":"10.00",
   "description":"My first payment",
   "method":null,
   "metadata":{  
      "order_id":"12345"
   },
   "details":null,
   "profileId":"pfl_HbC2MJ3mGj",
   "links":{  
      "paymentUrl":"https://www.mollie.com/payscreen/select-method/wbyGywdAPb",
      "webhookUrl":"https://webshop.example.org/payments/webhook/",
      "redirectUrl":"https://webshop.example.org/order/12345/"
   }
}
```

holtkamp:
Ah, sorry for your time spent on this. 

To be able to use IDE code completion / reduce typos etc, we also use an ```Mollie_API_Object_Payment``` object to assemble a request. Before sending it to the API it is converted to an array. 

But this is not the intended way of usage, my fault => closing!

Issue #423, opened 2019-11-26
TypeError when looping over issuers


JorikMollie:
## Specifications

  - API Version: 2.12.0

## Describe the issue

```
                     foreach ($method->issuers() as $issuer)
                        {
                        }

```
throws a TypeError with payment methods that do not have issuers.

sandervanhooft:
I'll have a look.

sandervanhooft:
PR prepared #424 .

I'll release as soon as I have received a "go".

sandervanhooft:
Released [in v2.12.1](https://github.com/mollie/mollie-api-php/releases/tag/v2.12.1)

PR #424, opened 2019-11-26
Fix empty subresource collection issue

sandervanhooft:
#423

### DIFF
--- a/src/Endpoints/MethodEndpoint.php
+++ b/src/Endpoints/MethodEndpoint.php
@@ -63,8 +63,8 @@ class MethodEndpoint extends CollectionEndpointAbstract
 
         return ResourceFactory::createBaseResourceCollection(
             $this->client,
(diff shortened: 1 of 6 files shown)


sandervanhooft:
I've expanded the scope to all calls made to `ResourceFactory::createBaseResourceCollection`.

sandervanhooft:
I'd prefer to set the array as a default value...

```php
class Method extends BaseResource {

public $issuers = [];

}
```

... but this could cause issues in existing integrations. So tackling this in  `ResourceFactory::createBaseResourceCollection` calls instead.

sandervanhooft:
@MollieRick Ready for release? (minor patch)

willemstuursma:
It might make more sense to update `createBaseResourceCollection` lest any callers forget to implement `?: []`.

sandervanhooft:
@willemstuursma Agreed. All done!

PR #422, opened 2019-11-21
Fix several hard errors

szepeviktor:
@sandervanhooft Hello! Please consider using @phpstan in CI.

This PR is inspired by `phpstan analyze -l 2 src/`

Level 4 is the minimum for a general project, if you handle money go with Level _max_.

This PR is written blindly, please correct me!

PR #426, opened 2019-11-29
Sneak peak into Level 2 problems...

szepeviktor:


szepeviktor:
Only 18 easy ones: https://travis-ci.org/mollie/mollie-api-php/jobs/618615266#L290

See https://github.com/mollie/mollie-api-php/pull/422#issue-344225626

Issue #438, opened 2020-01-24
Update metadata in subscriptions is not possible while docs say it should be


Cannonb4ll:
Hi,

I wanted to introduce a new feature in my system where users could easily update their subscription. However I am unable to update the 'metadata' inside a subscription.

If I check your code with the call in Subscription.php resource, it is indeed not send over.

```php
        $body = json_encode([
            "amount" => $this->amount,
            "times" => $this->times,
            "startDate" => $this->startDate,
            "webhookUrl" => $this->webhookUrl,
            "description" => $this->description,
            "mandateId" => $this->mandateId,
            "interval" => $this->interval,
        ]);
```

However, the docs clearly state it should be possible to send over new metadata:

https://docs.mollie.com/reference/v2/subscriptions-api/update-subscription#parameters

willemstuursma:
@sandervanhooft should be fixed in the Client?

sandervanhooft:
Just merged the fix. Will be released after the weekend.

sandervanhooft:
A fix was released with `v2.13.1` ([details](https://github.com/mollie/mollie-api-php/releases/tag/v2.13.1)).

Thanks for the PR and reporting @Cannonb4ll .

PR #439, opened 2020-01-25
Added metadata to update method

Cannonb4ll:
Referenced from issue #438


sandervanhooft:
Thanks @cannonb4ll, will be released after the weekend.

PR #444, opened 2020-02-11
Update PHPDoc for MethodPrice.php

pmzandbergen:
Fix for invalid PHPDoc in the MethodPrice class:

1. $description is a string, not a \stdClass object
2. $variable is a string representing a float, not a \stdClass object


willemstuursma:
Thanks for contributing!

PR #445, opened 2020-02-12
Update Payment.php

pmzandbergen:
Added missing response fields:

- amountCaptured
- applicationFeeAmount
- authorizedAt
- expiredAt (!= expiresAt)
- customerId
- restrictPaymentMethodsToCountry
- countryCode

Added functions:

- getAmountCaptured
- getApplicationFeeAmount

Suggestion: Return null instead of 0 when the requested amount object is unavailable (null).

### DIFF
--- a/src/Resources/Payment.php
+++ b/src/Resources/Payment.php
@@ -234,6 +234,58 @@ class Payment extends BaseResource
     public $isCancelable;
 
     /**
+     * The total amount that is already captured for this payment. Only available
+     * when this payment supports captures.
+     *
+     * @var \stdClass|null
+     */
+    public $amountCaptured;
+
+    /**
+     * The application fee, if the payment was created with one. Contains amount
+     * (the value and currency) and description.
+     *
+     * @var \stdClass|null
+     */
+    public $applicationFeeAmount;
+
+    /**
+     * The date and time the payment became authorized, in ISO 8601 format. This
+     * parameter is omitted if the payment is not authorized (yet).
+     *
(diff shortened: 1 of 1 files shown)


src/Resources/Payment.php
```diff
@@ -606,4 +658,34 @@ private function withPresetOptions(array $options)
...
+        return 0.0;
+    }
+
+    /**
+     * The total amount that is already captured for this payment. Only available
+     * when this payment supports captures.
+     * 
+     * @return float
+     */
+    public function getApplicationFeeAmount()
```
willemstuursma: I would prefer this to return the object with both the currency and the value.

pmzandbergen: Yes, however this would not match the corresponding functions like getAmountRemaining(). It won't make sense if one function returns a float while the other returns an object.

PS I don't use these functions at all, just the public properties:

``` php
$payment->setAmountRemaining(isset($resource->amountRemaining->value) ? (float) $resource->amountRemaining->value : null);
$payment->setAmountRemainingCurrency($resource->amountRemaining->currency ?? null);
```

sandervanhooft:
> Suggestion: Return null instead of 0 when the requested amount object is unavailable (null).

I'm ok with the current PR as-is: being consistent with existing amount getters. We can add some `hasX` methods later on if required.

sandervanhooft:
I'll schedule release for this week, probably Tuesday.

sandervanhooft:
Thanks @pmzandbergen !

PR #458, opened 2020-05-14
Add metadata variable to the order line resource

reindernl:
The metadata variable is missing from the resource. As you can see in the documentation provided by Mollie (https://docs.mollie.com/reference/v2/orders-api/create-order#order-lines-details), metadata can also be set on an order line. And it will be returned by Mollie if you fetch an order line through the API.

### DIFF
--- a/src/Resources/OrderLine.php
+++ b/src/Resources/OrderLine.php
@@ -125,6 +125,14 @@ class OrderLine extends BaseResource
      * @var string|null
      */
     public $productUrl;
+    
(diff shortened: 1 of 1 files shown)


sandervanhooft:
Could you check if it should be added to OrderLine.update() as well?

reindernl:
When I update an orderline with a metadata-param, everything works as expected.

sandervanhooft:
Thanks @reindernl !

PR #465, opened 2020-06-03
Fix profile endpoint test

Pimm:
HTTP status code 200 instead of 201. See https://docs.mollie.com/reference/v2/profiles-api/list-profiles


willemstuursma:
Cheers.

PR #467, opened 2020-06-17
Add dates to  order resource

edwwaarrdd:
Added missing date fields in Order.php
And added an extra assertion to the order endpoint test so that it is in line with the payment endpoint tests.
Took the dates from [https://docs.mollie.com/reference/v2/orders-api/get-order#response](https://docs.mollie.com/reference/v2/orders-api/get-order#response).

### DIFF
--- a/src/Resources/Order.php
+++ b/src/Resources/Order.php
@@ -145,6 +145,54 @@ class Order extends BaseResource
     public $createdAt;
 
     /**
+     * UTC datetime the order the order will expire in ISO-8601 format.
+     *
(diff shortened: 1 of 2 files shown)


edwwaarrdd:
There seems to be an error with the travis build when installing phpstan on php7.3.
`phpstan/phpstan 0.11.19 conflicts with webmozart/assert[1.9.0]`
I am not sure how to fix this.
Maybe @sandervanhooft you know more?

sandervanhooft:
I've opened a separate ticket for Travis / Phpstan (#474).

Thanks for this PR @edwwaarrdd !

PR #469, opened 2020-07-07
Added forId methods in SubscriptionEndpoint

Rkallenkoot:
Makes it consistent with the CustomerPaymentsEndpoint, saves a couple API calls when dealing with Subscriptions directly

### DIFF
--- a/src/Endpoints/SubscriptionEndpoint.php
+++ b/src/Endpoints/SubscriptionEndpoint.php
@@ -45,9 +45,7 @@ class SubscriptionEndpoint extends CollectionEndpointAbstract
      */
     public function createFor(Customer $customer, array $options = [], array $filters = [])
     {
-        $this->parentId = $customer->id;
-
-        return parent::rest_create($options, $filters);
+        return $this->createForId($customer->id, $options, $filters);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


sandervanhooft:
Thanks @Rkallenkoot ! Can you ensure it's covered by tests?

Rkallenkoot:
Sure, could you perhaps point out where `getFor` and such are tested so I have a point of reference?

sandervanhooft:
Easiest may be this approach (similar for other methods)

`getFor` is using `getForId` under the hood. If you refactor it that way, the existing tests for `getFor` will cover `getForId` as well.

sandervanhooft:
I.e. https://github.com/mollie/mollie-api-php/blob/65eb23e49725c13f9453d1c209d240d02add1be7/src/Endpoints/PaymentCaptureEndpoint.php#L44-L62

Rkallenkoot:
I'm not sure `getFor` etc are tested for Subscriptions btw!

Rkallenkoot:
Added methods are already covered :+1:

Rkallenkoot:
Let me know if I can assist in any way to get this merged!

sandervanhooft:
Thanks @Rkallenkoot ! Will be included in upcoming release.

Issue #472, opened 2020-07-09
Add API request parameters to Exception message if something goes wrong


willemstuursma:
We're often receiving lists of exception messages from our clients, for example:

```
[2020-06-04 10:19:54] app.CRITICAL: Mollie\Api\Exceptions\ApiException: cURL error 28: Operation timed out after 10001 milliseconds with 0 out of -1 bytes received (see http://curl.haxx.se/libcurl/c/libcurl-errors.html) (uncaught exception) at vendor/mollie/mollie-api-php/src/Exceptions/ApiException.php line 80 {"exception":"[object] (Mollie\\Api\\Exceptions\\ApiException(code: 0): cURL error 28: Operation timed out after 10001 milliseconds with 0 out of -1 bytes received (see http://curl.haxx.se/libcurl/c/libcurl-errors.html) at vendor/mollie/mollie-api-php/src/Exceptions/ApiException.php:80)"} []

[2020-06-04 15:32:21] app.CRITICAL: Mollie\Api\Exceptions\ApiException: cURL error 28: Operation timed out after 10000 milliseconds with 0 out of -1 bytes received (see http://curl.haxx.se/libcurl/c/libcurl-errors.html) (uncaught exception) at vendor/mollie/mollie-api-php/src/Exceptions/ApiException.php line 80 {"exception":"[object] (Mollie\\Api\\Exceptions\\ApiException(code: 0): cURL error 28: Operation timed out after 10000 milliseconds with 0 out of -1 bytes received (see http://curl.haxx.se/libcurl/c/libcurl-errors.html) at vendor/mollie/mollie-api-php/src/Exceptions/ApiException.php:80)"} []
...
```

It would be helpful if they had the original request attached with them as that would make it much more efficient to debug.

sandervanhooft:
This would probably only affect this exception throw:

https://github.com/mollie/mollie-api-php/blob/1eac11155cf048820941f7c3d94c429075195f6e/src/MollieApiClient.php#L458

sandervanhooft:
#473

PR #473, opened 2020-07-09
Add request to ApiException

sandervanhooft:
Adds a `getRequest` method to `ApiException`.

Solves #472

### DIFF
--- a/src/Exceptions/ApiException.php
+++ b/src/Exceptions/ApiException.php
@@ -2,8 +2,8 @@
 
 namespace Mollie\Api\Exceptions;
 
-use GuzzleHttp\Psr7\Request;
-use GuzzleHttp\Psr7\Response;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
 
 class ApiException extends \Exception
 {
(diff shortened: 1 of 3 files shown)


willemstuursma:
We should add the local time.

sandervanhooft:
> We should add the local time.

Sure. In Iso 8601 string or in php DateTime?

willemstuursma:
ISO is great. This PR is also missing the part where the request body is added to the exception message. 

The exception messages are often shared with us by merchants, so it is convenient if they contain more data to help track problems (so we don't have to ask for it).

sandervanhooft:
@willemstuursma Added both the dateTime iso string and the request body to the exception object and message.

Let me know if it can be merged and I'll release it.

willemstuursma:
@sandervanhooft I made some more changes. If you concur, it can be released.

sandervanhooft:
Thanks @willemstuursma , looks good to me! I'll schedule a release tomorrow.

lode:
This PR made it a little harder for us to recognize errors and act upon them.

Before this PR we checked the exception message. While maybe not perfect, it was the only thing we had which gave some information. With this PR the timestamp is added to that making it harder for us to auto determine what to do based on the message.

We now can use something like `json_decode($e->getResponse()->getBody()->__toString(), true)['detail']`. But it would be nice if there was a public interface on the exception object to get to the response json (or its actual fields) in a bit more stable way for which we don't have to know how the exception object works internally.

PR #475, opened 2020-07-15
Add paypal to valid first payment methods

mmachatschek:
As the API documentation states, it's possible to [create](https://docs.mollie.com/reference/v2/mandates-api/create-mandate) and [receive](https://docs.mollie.com/reference/v2/mandates-api/get-mandate) mandates with the method **paypal**.
Therefore this PR adds the payment method to the possible **MandateMethod** entries


sandervanhooft:
Big thanks (again!) @mmachatschek !

PR #497, opened 2020-10-01
BEL-770 Increase DNS cache timeout

willemstuursma:
This aims to reduce the number of DNS lookups needed in environments that support the global cURL DNS cache (e.g. most Apache configurations).


sandervanhooft:
LGTM. Perhaps best to test it first with the people experiencing these issues before merging into master?

willemstuursma:
I don't think we need this anymore.

Issue #498, opened 2020-10-07
Prevent requests timing out


sandervanhooft:
## Specifications

  - API Version:  v2

## Describe the issue

Lately there have been some reports about api requests timing out. Mollie's development teams (@willemstuursma @rares-mollie) expect their DNS to be a significant factor.

While the problem is being addressed at DNS / infrastructure level at Mollie, this issue addresses mitigations that can be implemented within the mollie-api-php client.

There are two approaches to this:

1. Tweaking the Guzzle client options.
2. Implementing a retry scheme on the Guzzle client.

I am mentioning both here because these approaches are interlinked.

### 1. Tweaking the Guzzle client options.

@rares-mollie has suggested to set the following on the Guzzle Options:

```bash
TIMEOUT: 10 # already there
CONNECT_TIMEOUT: 2 # new
READ_TIMEOUT: 10 # new
```

By setting the CONNECT_TIMEOUT to 2 we catch connection issues early on, before the call got through to the API. This ensures we are not getting timeout errors on a POST request that somehow still got through. Meaning we do not get false positive timeouts on calls that can mess things up in the API.

It also means that on `connect` issues, the retry scheme (when implemented) will kick in after 2 seconds instead of 10.

### 2. Implementing a retry scheme on the Guzzle client.

I'll look into applying a retry scheme to the Guzzle client using the `RetryMiddleware` that ships with Guzzle. Probably will require a factory and wrapper class.

This will impact `MollieApiClient::_construct()`, but it can be done in a backwards compatible way.

**When should the middleware be applied? (updated)**

- All requests (`GET`, `POST`, `PATCH`, `DELETE`) that result in a CONNECT timeout should be retried.
- All `GET` requests that connected ok but result in a timeout anyway should be retried.

**When should the middleware NOT be applied? (updated)**

- If there was no CONNECT timeout, all other requests should not use the retry middleware, especially the `POST` and `PATCH` requests.

**Testability**

- the modified Guzzle client, especially the `$decider` function which determines when the retry middleware should be applied
- this needs to be tested with Guzzle 6 and 7 in Travis
- this needs to work with PHP 5.6 and up

sandervanhooft:
@willemstuursma can you explain how #497 relates here, and if it still is considered part of the solution?

sandervanhooft:
Related: #492 (timeouts reported)

rares-mollie:
[...]

PR #499, opened 2020-10-08
Adding Guzzle Retry, modifying timeout settings

sandervanhooft:
See issue #498 

@rares-mollie @willemstuursma

### DIFF
--- a/src/MollieApiClient.php
+++ b/src/MollieApiClient.php
@@ -81,11 +81,6 @@ class MollieApiClient
     const CONNECT_TIMEOUT = 2;
 
     /**
-     * Default read timeout (in seconds).
-     */
-    const READ_TIMEOUT = 10;
-
-    /**
      * @var ClientInterface
      */
     protected $httpClient;
@@ -305,7 +300,6 @@ class MollieApiClient
                 GuzzleRequestOptions::VERIFY => CaBundle::getBundledCaBundlePath(),
                 GuzzleRequestOptions::TIMEOUT => self::TIMEOUT,
                 GuzzleRequestOptions::CONNECT_TIMEOUT => self::CONNECT_TIMEOUT,
-                GuzzleRequestOptions::READ_TIMEOUT => self::READ_TIMEOUT,
                 'handler' => $handlerStack,
             ]);
         }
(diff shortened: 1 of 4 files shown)


src/MollieApiClient.php
```diff
@@ -280,12 +294,21 @@ class MollieApiClient
...
+        if(! $this->httpClient) {
+            $retryMiddlewareFactory = new RetryMiddlewareFactory;
+            $handlerStack = HandlerStack::create();
+            $handlerStack->push($retryMiddlewareFactory->retry());
+
+            $this->httpClient = new Client([
+                GuzzleRequestOptions::VERIFY => CaBundle::getBundledCaBundlePath(),
+                GuzzleRequestOptions::TIMEOUT => self::TIMEOUT,
+                GuzzleRequestOptions::CONNECT_TIMEOUT => self::CONNECT_TIMEOUT,
+                GuzzleRequestOptions::READ_TIMEOUT => self::READ_TIMEOUT,
```
rares-mollie, line 308: The READ_TIMEOUT is applicable to stream reads which we don't use, even if we did we wouldn't be retrying the request as we only retry connect-time errors. I would recommend dropping READ_TIMEOUT and only using TIMOUT and CONNECT_TIMEOUT in order to avoid confusion over which is triggered.

sandervanhooft: Done

rares-mollie:
It looks good to me, @willemstuursma ?

sandervanhooft:
@rares-mollie One thing to consider is that currently the delay increases with 1000ms with each try. May make sense to lower it a bit, for example to 500ms.

rares-mollie:
I did consider this, my first thought was that the first retry should be quicker, but on second thought the incidence of triggering the retry strategy is quite rare and it splits into 2 cases:
 - sporadic: only about 1 in 5-10 x 1000 requests are affected
 - generalised: there's a L3-level incident affecting a large proportion of our clients

The first case has very limited impact and a delay is acceptable, in the second case the delay is preferred in order to avoid pile-ups of retries over very short periods of time. (eg: if you have immediate retry and you're down for a couple of seconds, you'll be handling all those requests being retry within the first <retry interval> of coming back up again) leading to a stampede

sandervanhooft:
Great, I'll go ahead and release then :)

PR #505, opened 2020-10-28
Add missing constant and fix orderLine::update()

SanderVanLeeuwen:
- Constant for "failed" refunds was missing
- OrderLine update didn't work when passing empty (null) parameters

### DIFF
--- a/src/Resources/OrderLine.php
+++ b/src/Resources/OrderLine.php
@@ -295,6 +295,20 @@ class OrderLine extends BaseResource
      */
     public function update()
     {
+        $url="orders/{$this->orderId}/lines/{$this->id}";
+        $body = json_encode($this->getData());
+        $result = $this->client->performHttpCall(MollieApiClient::HTTP_PATCH, $url, $body);
+
+        return ResourceFactory::createFromApiResult($result, new Order($this->client));
+    }
+
+    /**
+     * Get sanitized array of order line data
+     *
+     * @return array
+     */
+    public function getData()
+    {
         $data = array(
             "name" => $this->name,
             'imageUrl' => $this->imageUrl,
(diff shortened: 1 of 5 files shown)


sandervanhooft:
Hi @SanderVanLeeuwen ,

Nice! Could you provide a matching test for that filter?

SanderVanLeeuwen:
To be honest I don't have a lot of experience with writing tests. Not sure too how to simulate the `update()` call as a whole and if that's even possible. But would this code separation and test (see additional commit) be adequate? It feels like a lot of things aren't covered yet, but at least it does cover the newly added filter ;)

src/Resources/OrderLine.php
```diff
@@ -288,9 +288,28 @@ public function isSurcharge()
...
+
+        return ResourceFactory::createFromApiResult($result, new Order($this->client));
+    }
+
+    /**
+     * Get sanitized array of order line data
+     *
+     * @return array
+     */
+    public function getData()
```
sandervanhooft, line 310: Can you rename this to `getUpdateData`?

sandervanhooft (changes requested):
Hi @SanderVanLeeuwen ,

I think your approach is fine here. Only one little remark, after that it's ready for merge.

src/Resources/OrderLine.php
SanderVanLeeuwen: Done. I had named it like that originally but renamed it to `getData` because it can get the orderLine data regardless of updating or not. But this sounds reasonable as well ;)

sandervanhooft (approved):

sandervanhooft:
Thanks @SanderVanLeeuwen !

sandervanhooft:
Released in [v2.25.0](https://github.com/mollie/mollie-api-php/releases/tag/v2.25.0)

PR #507, opened 2020-11-02
Fixed PaymentMethodStatus::PENDING_ONBOARDING

Christian-Rades:
Hi, according to the [docs](https://docs.mollie.com/reference/v2/methods-api/get-method#parameters ) the status `PENDING_ONBOARDING` should have the value `pending-boarding`.


sandervanhooft:
On it

PR #514, opened 2020-11-30
Add php8 support and switch to Github Actions

ciungulete:
- Move tests from TravisCI to Github Actions
- Add PHPStan to Github Actions

Ping @sandervanhooft

PR #516, opened 2020-12-02
Add php8 support and switch to Github Actions

ciungulete:
Hi @sandervanhooft,

I think the pull request is ok now:
- [x] add php 8 support
- [x] migrate from Travis to GitHub Actions
- [x] use a syntax check to cover old php versions (<7.2) which are now incompatible with tests (phpunit introduced some mandatory signature changes, i.e. on setUp() and tearDown.
- [x] run phpscoper in actions

I cancel #514 because I made a mistake with git rebase 😖


sandervanhooft:
LGTM!

PR #519, opened 2020-12-14
Update PHP Scoper to 0.13.10 in Createzip.yml action

ciungulete:
@sandervanhooft, I update **PHP Scoper** to 0.13.10 in Createzip.yml action and run `php examples/initialize.php`. If the actions have errors won't create the zip file.


sandervanhooft:
Checked and confirmed! 🎉 

Thanks @ciungulete!

PR #520, opened 2020-12-16
Update .gitattributes

mvdnbrk:
We don't need the `examples` folder in production. 
Added `.php_cs` and `phpstan.neon` as well.


sandervanhooft:
Hi @mvdnbrk,

Thanks for the PR. A lot of developers have reported to enjoy the examples folder to be there, so we're leaving it in.

Issue #517, opened 2020-12-05
Some bugs/mistakes?


janssenr:
## Specifications

  - API Version: 2.0

## Describe the issue

In https://github.com/mollie/mollie-api-php/blob/df60ebfc2463cdffc1faeb1fa9419911812ac6f8/tests/Mollie/API/Endpoints/ShipmentEndpointTest.php "profileId" is used in getShipmentResponseFixture (row 488)
But in https://github.com/mollie/mollie-api-php/blob/master/src/Resources/Shipment.php I cannot find "profileId"

In https://github.com/mollie/mollie-api-php/blob/master/tests/Mollie/API/Endpoints/OrderPaymentEndpointTest.php "dueDate" is used (row 35 and 97)
But in https://github.com/mollie/mollie-api-php/blob/master/src/Resources/Payment.php I cannot find "dueDate"

In https://github.com/mollie/mollie-api-php/blob/master/tests/Mollie/API/Endpoints/SettlementEndpointTest.php "percentage" is used (row 79, 103, 127, 151, 296, 320, 344 and 368)
But in https://docs.mollie.com/reference/v2/settlements-api/get-settlement its called "variable"

In https://github.com/mollie/mollie-api-php/blob/master/tests/Mollie/API/Endpoints/OrderEndpointTest.php the following fields are used:
orderline.quantityShipped
orderline.amountShipped
orderline.quantityRefunded
orderline.amountRefunded
orderline.quantityCanceled
orderline.amountCanceled
orderline.shippableQuantity
orderline.refundableQuantity
orderline.cancelableQuantity
But in https://github.com/mollie/mollie-api-php/blob/master/src/Resources/OrderLine.php I cannot find those fields

In https://github.com/mollie/mollie-api-php/blob/master/tests/Mollie/API/Endpoints/OrderEndpointTest.php the fields "productUrl" and "imageUrl" are used within "_links" (row 326 and 380)
Is this correct? 

Sometimes api.mollie.nl is used. Change this to api.mollie.com?

sandervanhooft:
Hi @janssenr,

Everything is filled dynamically from the API response, but these resource fields can be helpful for autocompletion. Can you submit a PR?

PR #528, opened 2021-01-19
Add Missing fields

ciungulete:
Solve #517

### DIFF
--- a/src/Resources/Payment.php
+++ b/src/Resources/Payment.php
@@ -134,6 +134,25 @@ class Payment extends BaseResource
     public $failedAt;
 
     /**
+     * $dueDate is used only for banktransfer method
+     * The minimum date is tomorrow and the maximum date is 100 days after tomorrow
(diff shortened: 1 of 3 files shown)


src/Resources/Payment.php
```diff
@@ -133,6 +133,25 @@ class Payment extends BaseResource
      */
     public $failedAt;
 
+    /**
+     * $dueDate is used only for banktransfer method
```
sandervanhooft, line 137: Can you use this description instead? "The date the payment should expire. Please note: the minimum date is tomorrow and the maximum date is 100 days after tomorrow."

@example and @var are ok

sandervanhooft (changes requested):
One small change, otherwise ok!

src/Resources/Payment.php
sandervanhooft: Don't forget to pull in the latest master branch, I've made committed some changes

sandervanhooft (approved):

PR #538, opened 2021-03-09
[WIP]Guzzle extraction

sandervanhooft:
⚠️ Potential (0.01% chance) this PR includes a breaking change in frameworks that use some form of "autowiring". Please help me double check this before it gets merged. (See point 2 below).

----

**Reason**
Integrations like WooCommerce, Prestashop and OpenCart depend on this package. Their maintainers have reported a significant number of Guzzle collision issues over a longer period of time, even when using the pre-scoped edition of mollie-api-php.

The cleanest solution would be to remove the Guzzle dependency and any references to it. And introduce a MollieHttpAdapterInterface instead. However, this would cause a breaking change with too big an impact on integrations running in production. That's a no-go obviously.

**Proposed solution**
This PR proposes an hybrid way to avoid 99.99% of the impact of the change - but little impact can still occur:

1. Until this PR only `null` and `\GuzzleHttp\ClientInterface` were accepted on the `MollieApiClient` constructor. This PR adds a third option: a `MollieHttpAdapterInterface`.
1. To support the above the `ClientInterface` typehint on the constructor has been removed. Only in edge cases (the remaining 0.001%) where some form of customised "autowiring" is used to construct the `MollieApiClient` this can lead to issues. **Please check that your integration does not break if this `ClientInterface ` typehint ([link](https://github.com/mollie/mollie-api-php/blob/53f461ad6409a2d25a19d0336f88a285ade9b8e0/src/MollieApiClient.php#L291)) gets removed from the MollieApiClient constructor.**
1. The Guzzle dependency is now a _development_ dependency, to avoid collisions in production. Some version of Guzzle is included in most modern PHP frameworks.
1. A curl adapter is provided as a fallback if Guzzle 6 or 7 cannot be detected.
1. Some integrators have reported the need to support Guzzle 5. A Guzzle 5 adapter will not be included with this package, but it's easy to create your own and I don't expect any collisions here.

For many integrations this removes the need for downloading and integrating the pre-scoped build (a.k.a. "The Zip") on each upgrade. This has proven to be a brittle solution which I really hope to deprecate some time after this PR gets accepted.

TODO:

- [x] write the curl adapter
- [x] Fix some phpstan issues


sandervanhooft:
This PR will be tagged as a pre-release.

src/HttpAdapter/MollieHttpAdapterPicker.php
```diff
@@ -0,0 +1,60 @@
...
+
+        if ($httpClient instanceof MollieHttpAdapterInterface) {
+            return $httpClient;
+        }
+
+        if ($httpClient instanceof \GuzzleHttp\ClientInterface) {
+            return new Guzzle6And7MollieHttpAdapter($httpClient);
+        }
+
+        return new CurlMollieHttpAdapter;
```
lode, line 32: I think it would be nice to move returning the curl adapter to inside the `if (! $httpClient)` block, and to throw an exception here. Especially since the autowiring might get it wrong, but in general it feels weird that if the caller does something wrong we suddenly fallback to curl while we intended to use Guzzle.

sandervanhooft: Makes sense, thanks!

sandervanhooft: Here you go: https://github.com/mollie/mollie-api-php/pull/545#issue-610621456

Issue #437, opened 2020-01-21
Non-existent query parameter "testmode" for this API call


bobbybouwmann:
## Specifications
  - API Version: 2.12.1

## Describe the issue

I'm trying to revoke an existing mandate using the `revoke` method on the mandate resource I get back from the API. According to the documentation, the `testmode` parameter should be send along: https://docs.mollie.com/reference/v2/mandates-api/revoke-mandate#parameters

However, I get the following error back
```
"Error executing API call (422: Unprocessable Entity): Non-existent query parameter "testmode" for this API call.. Field: testmode. Documentation: https://docs.mollie.com/guides/handling-errors"
```

My code looks like this.
```
$mandates = $this->mollieApiClient->mandates->listForId($customerId, null, null, [
    'testmode' => $this->testmode,
]);

foreach ($mandates as $mandate) {
    $mandate->revoke();
}
```
Fetching the mandates works as expected. As you can see I call the `invoke` method on the Mandate resource. This method looks like this:
```
public function revoke()
{
    if (!isset($this->_links->self->href)) {
        return $this;
    }

    $body = null;
    if($this->client->L()) {
        $body = json_encode([
            "testmode" => $this->mode === "test" ? true : false
        ]);
    }

    $result = $this->client->performHttpCallToFullUrl(
        MollieApiClient::HTTP_DELETE,
        $this->_links->self->href,
        $body
    );

    return $result;
}
```
Should all be good, but I still get above mentioned error.

Any clues of what's going on?

sandervanhooft:
Can you dump the mandate contents here?

sandervanhooft:
Right now my best guess is the testmode 
 parameter should not be set on the request body (bug in revoke method). Also, it should already be included on the self link in the mandate response.

bobbybouwmann:
So the problem is that this is a `DELETE` route. Because of that, we need to post the `testmode` in the body. Like you said, the `?testmode=true`  is appended to the URL, which is not allowed according to the error message for this endpoint. 

Right now we have a URL that looks like this
```
https://api.mollie.com/v2/customers/cst_R9r23aF7Ku/mandates/mdt_Av5nsfq23G?testmode=true
```
And then with the `revoke` method, we DELETE the resource also appending the testmode stuff. 

So now, we need to strip the testmode=true from the URL ourselves. I can't imagine that we are the only one doing this and that there are no tests for this 🤔

[...]

PR #539, opened 2021-03-18
Workaround for testmode query param on cancel subscription call

ronaldtb:
Hi,

I've left a comment on #437 and this pull request contains a workaround for the `Error executing API call (422: Unprocessable Entity): Non-existent query parameter "testmode" for this API call.. Field: testmode` exception on the subscription->cancel() call.

In https://github.com/mollie/mollie-api-php/issues/437#issuecomment-580770762 it's mentioned that Mollie is accepting the `testmode` query parameter on the other endpoints, and I think adding this support also to the cancel subscription endpoint would be a better solution than merging this workaround.

### DIFF
--- a/src/Resources/Subscription.php
+++ b/src/Resources/Subscription.php
@@ -96,7 +96,7 @@ class Subscription extends BaseResource
      * @var \stdClass|null
      */
     public $webhookUrl;
-    
+
     /**
      * Date the next subscription payment will take place. For example: 2018-04-24
      *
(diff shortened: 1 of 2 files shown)


sandervanhooft:
Thanks @ronaldtb !

> In #437 (comment) it's mentioned that Mollie is accepting the testmode query parameter on the other endpoints, and I think adding this support also to the cancel subscription endpoint would be a better solution than merging this workaround.

I agree, I'll discuss this option with the Mollie team first

sandervanhooft:
Hang on @ronaldtb, we're still looking into this 😄

sandervanhooft:
Mollie will solve this in the API. I'll leave this PR open for reference until the issue has been solved.

ronaldtb:
Thanks for the update @sandervanhooft. I'm currently using the workaround from my pull request, so there's no hurry from my side.

Naoray:
seems like the issue was fixed

PR #547, opened 2021-04-19
Bugfix: Add missing return type

michielgerritsen:
The `\Mollie\Api\Endpoints\MandateEndpoint::getForId()` function is missing the correct return type, I've added it.


sandervanhooft:
Thanks @michielgerritsen !

Issue #549, opened 2021-04-23
Using Payment::listRefunds always causes an exception


johannesdr-tb:
## Specifications

  - API Version: 2

## Describe the issue

According to the API documentation, [you can pass a from and limit parameter when getting a list of all the refunds for a payment](https://docs.mollie.com/reference/v2/refunds-api/list-refunds#parameters).

When you try to do this with the API client `$payment->listRefunds($parameters)` this results in an exception "Error executing API call (400: Bad Request): Invalid value passed. Field: from."

Thee reason for this is that it's trying to pass the parameter array as the $from parameter to `EndpointAbstract::rest_list` here https://github.com/mollie/mollie-api-php/blob/master/src/Endpoints/PaymentRefundEndpoint.php#L87

I guess that either `Payment::listRefunds` needs to accept a $from and $limit parameter that gets passed along or it should extract those from the $parameters array.

sandervanhooft:
Thanks @johannesdr-tb for reporting this, I'll look into it.

sandervanhooft:
Fixed and released. Thanks @johannesdr-tb!

PR #550, opened 2021-04-24
Fix for payment refunds list exception

johannesdr-tb:
This is a fix for the issue described in #549. Adds missing from and limit parameters when calling rest_list to list all refunds for a payment.


sandervanhooft:
All good, thanks @johannesdr-tb !

PR #554, opened 2021-05-10
Change the PHPDoc for the subscription metadata

michielgerritsen:
The PHPDoc states that it contains an array, but it's actually a stdClass object.


sandervanhooft:
Good catch. Thanks @michielgerritsen !

PR #556, opened 2021-05-18
Add JSON content type to request

fjbender:
In https://github.com/mollie/mollie-api-php/blob/c1a500d251255e29919c6250d5c741ae4fe4e178/src/Endpoints/EndpointAbstract.php#L207-L211 we encode all data thrown into the endpoints as JSON. It would make sense to set an appropriate `Content-Type` header in this case, not only because it would be correct, but also because it eases handling when tunnelling API requests to Mollie through e.g. `ngrok` to analyze what exactly is happening.

Before:
image

After:
image

Caveats:
- ~~This hack also sets the `Content-Type` header for GET requests (e.g. `$mollie->payments->get('tr_12345')`), which is not explicitly forbidden but also not explicitly allowed in [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.5)~~
- I have not tested each and every request to Mollie

### DIFF
--- a/src/MollieApiClient.php
+++ b/src/MollieApiClient.php
@@ -482,8 +482,11 @@ class MollieApiClient
             'Accept' => "application/json",
             'Authorization' => "Bearer {$this->apiKey}",
             'User-Agent' => $userAgent,
-            'Content-Type' => "application/json",
         ];
+
+        if ($httpBody !== null) {
+            $headers['Content-type'] = "application/json";
+        }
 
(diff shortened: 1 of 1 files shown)


sandervanhooft:
Thanks @fjbender,

I've tested this with all possible REST calls against Mollie's API, and it seems to work fine.

Not sure about the merits of this PR though, I'll check with the API team.

PR #561, opened 2021-06-08
Add dueDate to update parameters 

Timollie:
Feature request from merchant to add this. Specifically for extending bank transfer windows. Perhaps filter on method an existence of dueDate is not needed.

### DIFF
--- a/src/Resources/Payment.php
+++ b/src/Resources/Payment.php
@@ -656,7 +656,8 @@ class Payment extends BaseResource
             "webhookUrl" => $this->webhookUrl,
             "metadata" => $this->metadata,
             "restrictPaymentMethodsToCountry" => $this->restrictPaymentMethodsToCountry,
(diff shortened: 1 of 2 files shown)


sandervanhooft:
That's very specific API logic to put into the client.

Perhaps always send the dueDate and have the API accept it for all methods - just ignore it when not applicable?

I'm also considering adding `$api->someEndpoint->update(array $params, array $filters);` methods. So you could pass in anything you want instead of relying on the `Resource.update()` to select the updatable fields for you.

Timollie:
Removed logic as it is not strictly required (tested).
Tests now fail on assumption the update JSON string should not contain dueDate.

A general update method for the endpoints would indeed be a way better solution.

Timollie:
Changed tests to expect dueDate in update call.

sandervanhooft:
Thanks @Timollie ! I'll include this in next week's release.

PR #562, opened 2021-06-09
NTR: add correct content type application/json

boxblinkracer:
content type was wrong.
It has a lower case "type" instead of "Type"

"Content-type" -> "Content-Type"

i did also create unit tests for the headers.
the guzzle though wasnt possible to mock with "->with" (always wrong numbers of parameters).
so i've created a small fake class, which is (at least i think) better anyway

i have 2 unit tests with 2 newables of that new client, i didnt want to screw up anything
by putting it into the global setUp...hope thats ok

### DIFF
--- a/src/MollieApiClient.php
+++ b/src/MollieApiClient.php
@@ -450,6 +450,10 @@ class MollieApiClient
             'User-Agent' => $userAgent,
         ];
 
+        if ($httpBody !== null) {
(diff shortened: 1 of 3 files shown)


sandervanhooft:
If it works I think it's perfect 😄

sandervanhooft:
Will be included in next week's release. Thanks @boxblinkracer !

PR #563, opened 2021-06-14
Add update method to PaymentEndpoint

ciungulete:
- Add `rest_update`  to `EndpointAbstract.php`
- Add example for update method
- Refactor Payment resource to use the new PaymentEndpoint.update() method

### DIFF
--- a/src/Resources/Payment.php
+++ b/src/Resources/Payment.php
@@ -646,10 +646,6 @@ class Payment extends BaseResource
 
     public function update()
     {
-        if (! isset($this->_links->self->href)) {
-            return $this;
-        }
-
         $body = [
             "description" => $this->description,
             "redirectUrl" => $this->redirectUrl,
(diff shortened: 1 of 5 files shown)


src/Resources/Payment.php
```diff
@@ -649,21 +649,18 @@ public function update()
...
             "metadata" => $this->metadata,
             "restrictPaymentMethodsToCountry" => $this->restrictPaymentMethodsToCountry,
+            "locale" => $this->locale,
             "dueDate" => $this->dueDate,
-        ]);
+        ];
 
-        $result = $this->client->performHttpCallToFullUrl(
-            MollieApiClient::HTTP_PATCH,
-            $this->_links->self->href,
```
sandervanhooft, line 664: This resource link included the `testmode` parameter when applicable. So if testmode was enabled to obtain the Payment, it would be preset on the `$payment->update()` call. Can you find a way to keep this behaviour?

Note that when testmode is enabled: `$payment->mode === 'test'`

I think this behaviour is already present on some resource calls, something with "`preset`" in the method name if I recall correctly.

sandervanhooft (changes requested):

sandervanhooft (approved):

PR #575, opened 2021-08-31
Implement route payments

ciungulete:


sandervanhooft (changes requested):
Comments may be outdated now...

src/Resources/Route.php
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Amount object containing the value and currency
+     *
+     * @var \stdClass
+     */
+    public $amount;
+
+    /**
+     * The destination where the routed payment was send.
```
sandervanhooft, line 27: senT 😄

tests/Mollie/API/Endpoints/PaymentRouteEndpointTest.php
```diff
@@ -0,0 +1,141 @@
+<?php
+
+namespace Tests\Mollie\Api\Endpoints;
+
+use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\Psr7\Response;
+use Mollie\Api\Resources\Payment;
+
+class PaymentRouteEndpointTest extends BaseEndpointTest
```
sandervanhooft, line 9: Only need to test updating the payment route here.

What's being tested here is already / should be covered by the `PaymentEndpointTest`.

src/Resources/Payment.php
```diff
@@ -276,6 +276,16 @@ class Payment extends BaseResource
      */
     public $applicationFee;
 
+    /**
+     * An optional routing configuration which enables you to route a successful payment,
+     * or part of the payment, to one or more connected accounts. Additionally, you can
+     * schedule (parts of) the payment to become available on the connected account on a
+     * future date.
+     *
+     * @var \array|null
```
sandervanhooft, line 285: Can you double check that an actual `array` is being built from the response, and not an object?

sandervanhooft (approved):

sandervanhooft:
@ciungulete can you include these changes too in this PR?

image

([From the changelog](https://docs.mollie.com/changelog/v2/changelog))

PR #577, opened 2021-09-07
Add missing @throws tags

mikey242:
Adds @throws tags to various method of the Customer resource class. Should make it easier for devs to know when to expect a possible exception.


sandervanhooft:
Hi @mikey242 ! Thanks for the effort. Looks like @ciungulete just beat you to it with #576 🤦 .

PR #581, opened 2021-10-05
Change type of Invoice.lines

holtkamp:
Probably the documentation at https://docs.mollie.com/reference/v2/invoices-api/get-invoice should also be updated


sandervanhooft:
Thanks @holtkamp !

PR #588, opened 2021-10-29
Added `getAmountChargedBack` to payment resource

edwinvdpol:
I noticed that the other amounts had this kind of method, but it was missing for the charged back amount.

### DIFF
--- a/src/Resources/Payment.php
+++ b/src/Resources/Payment.php
@@ -497,6 +497,21 @@ class Payment extends BaseResource
     }
 
     /**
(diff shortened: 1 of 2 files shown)


sandervanhooft:
Good catch @edwinvdpol! Can you add an accompanying test to `PaymentTest`?

edwinvdpol:
I added the tests. Please let me know if I missed anything.

sandervanhooft:
LGTM! Thanks!

PR #589, opened 2021-11-04
Add new partner api endpoints

ciungulete:


sandervanhooft:
Can you add this related one too?

https://docs.mollie.com/reference/v2/organizations-api/get-partner

sandervanhooft:
And can you add a `PartnersEndpointTest`?

PR #607, opened 2022-03-11
Drop Php 5.6

ciungulete:


sandervanhooft:
@ciungulete thanks!

- There's an error in the composer.json, missing `,` as far as I can see.
- 5.6 can also be removed here: https://github.com/mollie/mollie-api-php/blob/26010bd6999af4466c31c92733df87dc04d95772/.github/workflows/check_php_syntax.yml#L13

PR #615, opened 2022-04-08
Adding reason and reversedAt attributes to Chargeback resource

dmdijkslag:
Mollie has 2 additional attributes for its chargebacks endpoint, see https://docs.mollie.com/reference/v2/chargebacks-api/get-payment-chargeback for more info.

This will update the Chargeback resourse with the additional attributs.

For possible reason codes see https://help.mollie.com/hc/nl/articles/115000309865

### DIFF
--- a/src/Resources/Chargeback.php
+++ b/src/Resources/Chargeback.php
@@ -54,6 +54,14 @@ class Chargeback extends BaseResource
     public $reason;
 
     /**
+     * UTC datetime the date and time the chargeback was reversed in ISO-8601 format.
+     *
+     * @example "2013-12-25T10:30:54+00:00"
+     * @var string|null
+     */
+    public $reversedAt;
+
+    /**
      * @var \stdClass
      */
     public $_links;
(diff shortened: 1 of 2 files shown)


sandervanhooft:
Thanks @dmdijkslag ! You're right. Only one remark:

src/Resources/Chargeback.php
```diff
@@ -46,6 +46,21 @@ class Chargeback extends BaseResource
      */
     public $settlementAmount;
 
+    /**
+     * The chargeback reason
+     *
+     * @var \stdClass
```
sandervanhooft, line 52: `reason` can be `null`. It's only included on `directdebit` transactions

sandervanhooft (changes requested):
reason can be `null`

dmdijkslag:
@sandervanhooft thanx for the update, you are right. 
I updated the resource and test.

sandervanhooft (approved):

sandervanhooft:
Thanks @dmdijkslag !

PR #633, opened 2022-07-23
Improve static analyses from Phpstan to lvl5

ciungulete:


src/Endpoints/SettlementPaymentEndpoint.php
```diff
@@ -34,7 +33,7 @@ protected function getResourceCollectionObject($count, $_links)
      * @param int $limit
      * @param array $parameters
      *
-     * @return BaseCollection|PaymentCollection
+     * @return mixed
```
sandervanhooft, line 36: `PaymentCollection`

src/HttpAdapter/Guzzle6And7MollieHttpAdapter.php
```diff
@@ -101,10 +101,6 @@ public function send($httpMethod, $url, $headers, $httpBody)
             throw new ApiException($e->getMessage(), $e->getCode(), null, $request, null);
         }
 
-        if (! $response) {
```
sandervanhooft, line 104: Why was this removed?

sandervanhooft (changes requested):
Going in the right direction!

Don't forget that the reason to start this PR are especially these docblocks here:

https://github.com/mollie/mollie-api-php/blob/43ae5471967a47b34752b6b3a229038a05034527/src/Resources/BaseResource.php#L8-L37

We should find a way to remove these.

PR #635, opened 2022-08-07
Fix PhpStan Tests

ciungulete:


sandervanhooft:
Hi @ciungulete,

Is there another way to fix this? This solution would essentially devaluate the use of phpstan.

ciungulete:
Yes! Now it's done

sandervanhooft:
Looks great but it seems it's not running the full GitHub action suite on this PR?

sandervanhooft:
Thanks @ciungulete ! 🔥

PR #651, opened 2022-12-11
Fixed call to undefined function in example list payment links

reboot1983:
Fixed call to undefined function in example list payment links


sandervanhooft (approved):
Good catch!

PR #659, opened 2022-12-29
Add cancelUrl to Payment and Order resources

sandervanhooft:


sandervanhooft:
Awaiting Mollie API behaviour confirmation from @mollierobbert (no rush)

mollierobbert:
Last week we fixed the issue where you could not update the `cancelUrl` from some value to an empty value. That was the issue you were still keeping this open for, right? Or are we still missing something?

sandervanhooft:
@mollierobbert Especially from `null` to `null`. Seems to work! Merging now.

Thanks!

PR #665, opened 2023-03-09
Bugfix: Add metadata property

michielgerritsen:
When creating a refund, we would receive this error in PHP 8.2:

```php
Deprecated Functionality: Creation of dynamic property Mollie\Api\Resources\Refund::$metadata 
```

This PR adds the property to the class, although I'm not 100% sure about the type hint.


sandervanhooft:
Released! Thanks @michielgerritsen !

PR #670, opened 2023-03-28
Bugfix: Add resource property

bobvandevijver:
Basically the same as https://github.com/mollie/mollie-api-php/pull/665.

When loading the issuers, we would receive this error in PHP 8.2:

```
Deprecated: Creation of dynamic property Mollie\Api\Resources\Issuer::$resource is deprecated
```

This PR adds the property to the class.

### DIFF
--- a/src/Resources/Issuer.php
+++ b/src/Resources/Issuer.php
@@ -12,6 +12,11 @@ class Issuer extends BaseResource
     public $id;
 
     /**
+     * @var string
+     */
+    public $resource;
+
+    /**
      * Name of the issuer.
      *
      * @var string


sandervanhooft (approved):

bobvandevijver:
Okay, so this was my fault, I'm sorry for that. Apparently I based my changes on a outdated release due to not having the latest one installed as Renovate has apparently failed to update this dependency for quite a while. MR is fixed now!

sandervanhooft:
Reversion done.

bobvandevijver:
You're right, I was confused as `Method` does define the resource separately. I will be making another PR to remove all those duplicates.

bobvandevijver:
Although, it might be better to add a full description to the docblock for this particular class

sandervanhooft:
Yeah that's a bit inconsistent. Looking forward to your PR :)

PR #673, opened 2023-03-30
Cleanup unneeded property overrides from Resource classes

bobvandevijver:
I've kept the resource properties in the classes that actually add a description.

### DIFF
--- a/src/Resources/Capture.php
+++ b/src/Resources/Capture.php
@@ -6,6 +6,7 @@ class Capture extends BaseResource
 {
     /**
      * Always 'capture' for this object
+     *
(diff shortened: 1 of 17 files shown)


Naoray:
Hi @bobvandevijver,

Thanks for your contribution! However, since every endpoint returns a `resource` string, we still need the property in every resource class.

bobvandevijver:
@Naoray I know, that is why I only removed the unneeded overrides of the base class. It is still defined in https://github.com/mollie/mollie-api-php/blob/master/src/Resources/BaseResource.php#L22.

Issue #668, opened 2023-03-24
RefundStatus::STATUS_CANCELED is missing


Semvrij:
## Describe the issue

RefundStatus::STATUS_CANCELED is missing.

https://docs.mollie.com/payments/refunds#refund-statuses

Naoray:
Hi @Semvrij,

could you provide us with more information on what you mean by that please?

erikkraijenoord:
@Naoray 
I think he is refering to the file not having the `STATUS_CANCELED` but is documented in the docs:
```php
namespace Mollie\Api\Types;

class RefundStatus
{
    /**
     * The refund is queued until there is enough balance to process te refund. You can still cancel the refund.
     */
    public const STATUS_QUEUED = 'queued';

    /**
     * The refund will be sent to the bank on the next business day. You can still cancel the refund.
     */
    public const STATUS_PENDING = 'pending';

    /**
     * The refund has been sent to the bank. The refund amount will be transferred to the consumer account as soon as possible.
     */
    public const STATUS_PROCESSING = 'processing';

    /**
     * The refund amount has been transferred to the consumer.
     */
    public const STATUS_REFUNDED = 'refunded';

    /**
     * The refund has failed after processing. For example, the customer has closed his / her bank account. The funds will be returned to your account.
     */
    public const STATUS_FAILED = 'failed';
}

PR #677, opened 2023-05-03
Adding Canceled status to \Types\RefundStatus

vannut:
As per #668


Naoray:
@vannut Thanks for your contribution, I’ve added another method to the `Resources\Refund` and added it to your commit in #678

PR #684, opened 2023-05-18
Update PaymentLink.php add archived parameter

bagulho:
## Specifications

  - API Version: v2
  - Endpoint: get-payment-link

## Issue

The get-payment-link endpoint defines an `archived` field in the response, but the related class `vendor/mollie/mollie-api-php/src/Resources/PaymentLink.php` doesn't have the attribute.


Naoray:
Thanks for the contribution @bagulho!

PR #687, opened 2023-06-19
Add create client-links endpoint

Naoray:


src/Resources/ClientLink.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        if (! in_array($approval_prompt, [ApprovalPrompt::AUTO, ApprovalPrompt::FORCE])) {
+            throw new \Exception('Invalid approval_prompt. Please use "auto" or "force".');
+        }
+
+        $query = http_build_query([
+            'client_id' => $client_id,
+            'state' => $state,
+            'approval_prompt' => $approval_prompt,
+            'scopes' => implode('+', $scopes),
```
lucasfeliciano, line 44: There is a typo in our documentation, it should be `scope` instead of `scopes`. 
I'll also fix the docs.

Also, I had to change the character to implode, in the example in the documentation we have the encoded URL.
`http_build_query` encodes space as `+`. 

```suggestion
            'scope' => implode(' ', $scopes),
```

src/Resources/ClientLink.php
```diff
@@ -0,0 +1,49 @@
...
+     * @var \stdClass
+     */
+    public $_links;
+
+    /**
+     * Get the redirect URL where the customer can complete the payment.
+     *
+     * @return string|null
+     */
+    public function getRedirectUrl(string $client_id, string $state, string $approval_prompt, array $scopes = [])
```
lucasfeliciano, line 34: We don't require `approval_prompt`, instead we have `auto` as default.

```suggestion
    public function getRedirectUrl(string $client_id, string $state, array $scopes = [], string $approval_prompt= 'auto')
```

src/Endpoints/ClientLinkEndpoint.php
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Creates a client link in Mollie.
+     *
+     * @param array $data An array containing details on the client link.
+     * @param array $filters
+     *
+     * @return ClientLink
+     * @throws ApiException
+     */
+    public function create(array $data = [], array $filters = []): ClientLink
```
lucasfeliciano, line 37: Genuine question:

Why the `$filters` parameter? I don't think it is used at all.

lucasfeliciano (changes requested):
Hi, thank you so much for this PR.

I've tested and I found some minor details. Mostly because wrong information in our documentation.

I'll make sure to have it fixed on our side.

src/Endpoints/ClientLinkEndpoint.php
sandervanhooft: @lucasfeliciano Good catch. It's there for consistency with other endpoints mostly. It can be removed. But in 90% of the cases I see it then getting added within the year. What do you prefer?

lucasfeliciano: Thanks for clarifying @sandervanhooft 

I would rather remove to provide a good developer experience.
It could be confusing to pass a filter and don't see it being reflected in the response

If we need in the future, we can add it. 🙏

sandervanhooft: Agreed!

PR #688, opened 2023-06-20
Cleanup unneeded property overrides from Resource classes

bobvandevijver:
Resubmit of #673 which has been closed incorrectly without further response. This change removes unneeded, duplicated overrides, the `$resource` property already resides in the `BaseResource` class.

### DIFF
--- a/src/Resources/Capture.php
+++ b/src/Resources/Capture.php
@@ -6,6 +6,7 @@ class Capture extends BaseResource
 {
     /**
      * Always 'capture' for this object
+     *
      * @var string
      */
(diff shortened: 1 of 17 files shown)


Naoray:
Thank you @bobvandevijver for your submission. To be clear: I did plan on reopening your older PR. Sorry for not reopening it sooner, I had no bad intentions ❤️

bobvandevijver:
That's okay, I just assumed that the notification might have been lost in the message stream and having an open PR is just a bit more visible :)

PR #697, opened 2023-10-09
Run CI tests on PHP 8.3

jnoordsij:
With a few RC's released and the final release coming up in november, it might be a good thing to start running tests on the new PHP.


sandervanhooft:
Thanks @jnoordsij !

PR #698, opened 2023-10-10
Add constant for Klarna One method.

markbeazley:
Add a constant for the new Klarna one method that's used to provide Klarna in the UK.

Wasn't able to add a link as I don't think you have a public URL for the method yet as its still in beta.


sandervanhooft:
Awaiting info and confirmation from the Mollie team.

sandervanhooft:
Thanks @markbeazley !

PR #702, opened 2023-10-30
Feature/add auto pagination helper methods

Naoray:
This is a follow up PR of #701 to add helper methods to the auto pagination functionality. Similar to the advanced usage of the [nodejs implementation](https://github.com/mollie/mollie-api-node/pull/235).

In general this PR allows to chain methods onto the `getAutoPaginator()` method call in order to manipulate items within the collection. The available methods are

- `all()` -> returns the current page items as array
- `get()` -> enables to receive a specific item from the collection
- `filter()` -> filters through all entries and generates a new collection from the filtered result
- `first()` -> returns first item of the collection passing the given truth test
- `map()` -> returns the mapped collection for the given callback
- `take()` -> returns a new collection with only X items
- `every()` -> determines if all items of the collection pass a given truth test
- `count()` -> receive the count of the current collection items

This new approach allows for the following usages

```php
$iterator = $client->payments
    ->iterator()
    ->filter(fn (Payment $payment) => $payment->amount->currency === ‘€’ && $payment->amount->value > 100)
    ->take(10);

foreach ($iterator as $payment) {
    //…
}
```


src/Resources/LazyCollection.php
```diff
@@ -0,0 +1,169 @@
...
+                yield $iterator->key() => $iterator->current();
+
+                if ($limit) {
+                    $iterator->next();
+                }
+            }
+        });
+    }
+
+    public function every(callable $callback): bool
```
sandervanhooft, line 129: What does this do? docblock?

src/Resources/LazyCollection.php
```diff
@@ -0,0 +1,169 @@
+<?php
+
+namespace Mollie\Api\Resources;
+
+use ArrayIterator;
+use IteratorAggregate;
+use Traversable;
+
+class LazyCollection implements IteratorAggregate
```
sandervanhooft, line 9: In cases this LazyCollection gets returned, are we loosing access to any methods available on the original concrete collection classes?

src/Resources/LazyCollection.php
Naoray: It checks if every item in the collection pass the given truth test. Added the docblock

src/Resources/LazyCollection.php
Naoray: @sandervanhooft Yes, we loose all access to ‘standard’ collection methods. I’ve check the other mollie-api sdks and none of them have custom collection methods per resources.

The only idea that I have on how to make this work is to allow a Collection class to define a dedicated LazyCollection class which is then returned.

```php
class MandateCollection extends CursorCollection
{
    public function getLazyCollectionName(): string
    {
        return LazyMandateCollection::class;
    }

    //…
}

class LazyMandateCollection extends LazyCollection
{
    public function whereStatus($status)
    {
        return $this->filter(fn (Mandate $mandate) => $mandate->status === $status);
    }
}

class CursorCollection extends BaseCollection
{
    //...

    public function getAutoIterator(bool $iterateBackwards = false): LazyCollection
    {
        $page = $this;

        $customCollection = $page->getLazyCollectionName();

        $iteratorFunction = function () use ($page, $iterateBackwards): Generator {
            while (true) {
                foreach ($page as $item) {
                    yield $item;
                }

                if (($iterateBackwards && !$page->hasPrevious()) || !$page->hasNext()) {
                    break;
                }

                $page = $iterateBackwards
                    ? $page->previous()
                    : $page->next();
            }
        };

        return $customCollection
            ? new $customCollection($iteratorFunction)
            : new LazyCollection($iteratorFunction);
    }
}
```

PR #704, opened 2023-12-19
NTR: Update return.php to include sync poll from API

fjbender:
We should make it a bit more clear that it is a possibility that the webhook has not yet arrived when the consumer returns.

See point 6 here: https://docs.mollie.com/payments/accepting-payments#working-with-the-payments-api


sandervanhooft (approved):

sandervanhooft:
Thanks!

PR #716, opened 2024-05-01
Fix typo

iDiegoNL:
Fixes a random typo that I noticed :)


fjbender:
Thank you for your contribution! It ain't much, but it's honest work!

PR #724, opened 2024-06-21
Add parallel testing and drop abandoned liberator package

sandervanhooft:


Naoray:
@sandervanhooft that breaks compatibility with php < 7.3

sandervanhooft:
That's a pity. Let's keep it on the back burner.

PR #735, opened 2024-07-10
Enhanced methods.allActive() for integration consistency and added methods.list()

sandervanhooft:
This PR ensures integration consistency: `methods.allActive()` now returns the full set of activated methods regardless the currencies supported.

Context: Mollie added support for `twint` recently, which does not appear on `/v2/methods` by default because Twint does not support the `EUR` currency.

To mirror the API, the `list()` method is added for directly interacting with `v2/methods`.

```php
$mollie->methods->list(); // Directly interact with the "/v2/methods" endpoint
$mollie->methods->allActive(); // Enhanced for consistency: returns all methods regardless the currencies supported
```

### DIFF
--- a/src/Endpoints/MethodEndpoint.php
+++ b/src/Endpoints/MethodEndpoint.php
@@ -45,12 +45,15 @@ class MethodEndpoint extends CollectionEndpointAbstract
      */
     public function allActive(array $parameters = [])
     {
-        $url = 'methods/all' . $this->buildQueryString($parameters);
+        $url = "methods/all" . $this->buildQueryString($parameters);
+        $result = $this->client->performHttpCall("GET", $url);
 
-        $result = $this->client->performHttpCall('GET', $url);
+        $allowedStatuses = $this->isUsingTestmode($parameters)
+            ? [PaymentMethodStatus::ACTIVATED, PaymentMethodStatus::PENDING_REVIEW]
+            : [PaymentMethodStatus::ACTIVATED];
 
-        $data = array_filter($result->_embedded->methods, function ($method) {
-            return $method->status === PaymentMethodStatus::ACTIVATED;
+        $data = array_filter($result->_embedded->methods, function ($method) use ($allowedStatuses) {
+            return in_array($method->status, $allowedStatuses);
         });
 
         unset($result->_links->self); // Remove incorrect self link
@@ -129,5 +132,25 @@ class MethodEndpoint extends CollectionEndpointAbstract
     public function list(array $parameters = [])
     {
         return $this->rest_list(null, null, $parameters);
+    }
+
+    /**
+     * @param array $parameters
+     * @return bool
+     */
+    protected function isUsingTestmode(array $parameters): bool
+    {
+        if ($this->client->usesOAuth()) {
+            if (
+                array_key_exists("testmode", $parameters)
+                && is_bool($parameters["testmode"])
+            ) {
+                return $parameters["testmode"];
+            }
+
+            return false;
+        }
+
+        return $this->client->usesTestmodeApiKey();
     }
 }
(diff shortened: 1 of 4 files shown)


src/Endpoints/MethodEndpoint.php
```diff
@@ -44,7 +45,22 @@ public function all(array $parameters = [])
...
     public function allActive(array $parameters = [])
     {
-        return parent::rest_list(null, null, $parameters);
+        $url = 'methods/all' . $this->buildQueryString($parameters);
+
+        $result = $this->client->performHttpCall('GET', $url);
+
+        $data = array_filter($result->_embedded->methods, function ($method) {
+            return $method->status === PaymentMethodStatus::ACTIVATED;
+        });
```
fjbender, line 54: `PaymentMethodStatus::ACTIVATED` is for live mode, in test mode `PaymentMethodStatus::ACTIVATED` and the `PENDING` status should be considered. See https://arc.net/l/quote/pgdugzxj

tests/Mollie/API/Endpoints/MethodEndpointTest.php
```diff
@@ -312,28 +311,23 @@ public function testGetTranslatedMethod()
...
-
-        $documentationLink = (object)[
-            'href' => 'https://docs.mollie.com/reference/v2/methods-api/get-method',
-            'type' => 'text/html',
-        ];
+        $this->assertLinkObject(
+            'https://api.mollie.com/v2/methods/sofort',
+            'application/hal+json',
+            $method->_links->self
+        );
```
fjbender, line 318: As `sofort` has been deprecated, maybe we should consider updating this test :)

tests/Mollie/API/Endpoints/MethodEndpointTest.php
```diff
@@ -641,4 +631,160 @@ public function testListAllAvailableMethods()
             $creditcardMethod->_links->self
         );
     }
+
+    public function testListMethods()
+    {
+        $this->mockApiCall(
+            new Request('GET', '/v2/methods'),
+            new Response(
```
fjbender, line 639: Aren't we missing the `status` here, too?

fjbender (changes requested):

fjbender:
> `methods.allActive()` now returns the full set of activated methods regardless the currencies supported.

Should we consider this a "soft" BC-break? Or is the filter behavior retained when we specify e.g. amount and currency filter conditions?

tests/Mollie/API/Endpoints/MethodEndpointTest.php
sandervanhooft: Done, switched to creditcard (Karte)

tests/Mollie/API/Endpoints/MethodEndpointTest.php
sandervanhooft: :+1: Added to stubs for consistency.

src/Endpoints/MethodEndpoint.php
sandervanhooft: @fjbender Which `pending` statuses should I check for? Assuming only `pending-review`?

image

src/Endpoints/MethodEndpoint.php
```diff
@@ -44,7 +45,25 @@ public function all(array $parameters = [])
      */
     public function allActive(array $parameters = [])
     {
-        return parent::rest_list(null, null, $parameters);
+        $url = "methods/all" . $this->buildQueryString($parameters);
+        $result = $this->client->performHttpCall("GET", $url);
```
Naoray, line 49: why not use the existing `allAvailable()` method here which already fetches the `methods/all` endpoint

```suggestion
        $result = $this->allAvailable($parameters);
```

sandervanhooft:
Closing this for now

PR #746, opened 2024-10-13
fix: wrong phpdoc return for cancelFor/cancelForId methods

mheiduk:
the return for the subscription methods cancelFor/cancelForId in phpdoc comments is wrong. instead of null it should be mixed.

see https://docs.mollie.com/reference/cancel-subscription

### DIFF
--- a/src/Endpoints/SubscriptionEndpoint.php
+++ b/src/Endpoints/SubscriptionEndpoint.php
@@ -194,7 +194,7 @@ class SubscriptionEndpoint extends CollectionEndpointAbstract
      * @param string $subscriptionId
      * @param array $data
      *
-     * @return mixed
+     * @return Subscription
      * @throws ApiException
      */
     public function cancelFor(Customer $customer, $subscriptionId, array $data = [])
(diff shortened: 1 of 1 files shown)


sandervanhooft:
Hi @mheiduk ,

Thanks for the PR.

On this package the return value actually is `null` here.

mheiduk:
With the `cancelForId` method I get a response like the documentation says:

```php
$subscription = $this->mollie->subscriptions->cancelForId($customer_id, $subscription_id);
return rest_ensure_response(['status' => $subscription->status]);
```

response is in this case:

```json
{
  "status": "canceled"
}
```

mheiduk:
push @sandervanhooft 

would be great if you can take a second look at this.

sandervanhooft:
@mheiduk turns out you're right! 😅  Sorry about that.

sandervanhooft:
I'd prefer to change the return type to `Subscription` here though.

sandervanhooft:
`Mollie\Api\Resources\Subscription`

mheiduk:
@sandervanhooft thanks, changed it to Subscription

sandervanhooft:
Thanks @mheiduk, released it just now!

https://github.com/mollie/mollie-api-php/releases/tag/v2.74.0

PR #756, opened 2024-12-10
Feature/add sales invoices endpoint

Naoray:


examples/sales-invoices/create-sales-invoice.php
```diff
@@ -0,0 +1,51 @@
...
+
+    /*
+     * Create a sales invoice
+     */
+    $salesInvoice = $mollie->salesInvoices->create([
+        'currency' => 'EUR',
+        'status' => SalesInvoiceStatus::DRAFT,
+        'vatScheme' => 'standard',
+        'vatMode' => 'inclusive',
+        'paymentTerm' => PaymentTerm::DAYS_30,
```
sandervanhooft, line 24: This Type does not exist, perhaps for the better.

phpstan-baseline.neon
```diff
@@ -220,6 +220,31 @@ parameters:
 			count: 1
 			path: examples/profiles/update-profile.php
 
+		-
+			message: "#^Access to constant DAYS_30 on an unknown class Mollie\\\\Api\\\\Types\\\\PaymentTerm\\.$#"
```
sandervanhooft, line 224: Phpstan picked up on it too

sandervanhooft (changes requested):

examples/sales-invoices/create-sales-invoice.php
Naoray: I don't think having this type d be bad.

sandervanhooft: Let's add it :+1:

src/Types/SalesInvoiceStatus.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Mollie\Api\Types;
+
+class SalesInvoiceStatus
+{
+    /**
+     * The sales invoice is in draft status and has not been sent or paid.
+     */
+    const DRAFT = 'draft';
```
sandervanhooft, line 10: These constants can be marked as `public`

sandervanhooft (approved):

PR #759, opened 2025-01-20
V3 without dedicated payloads and queries

Naoray:


sandervanhooft:
Loving these fat requests!

PR #760, opened 2025-01-24
Endpoint ApiException message fix

MrTieGuy06:
Hello there!

While using the this API, I found that two of the Endpoint classes had an incorrect error message.
I also checked the other endpoint classes and cleaned up minor mistakes.


sandervanhooft:
Good catch!

PR #762, opened 2025-02-06
Feature/add capabilities endpoint

Naoray:


src/Endpoints/CapabilityEndpoint.php
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Retrieve all capabilities.
+     *
+     * @param array $parameters
+     *
+     * @return CapabilityCollection
+     * @throws ApiException
+     */
+    public function list(array $parameters = [])
```
sandervanhooft, line 44: Unpaginated, on v2 we have been using `all()` for these. I recommend to do the same here.

sandervanhooft (changes requested):
Small change requested. Also consider introducing a dedicated CapabilityStatusType.

PR #770, opened 2025-03-11
Add MB Way and Multibanco to PaymentMethod type

mweghorst:


src/Types/PaymentMethod.php
```diff
@@ -122,6 +122,16 @@ class PaymentMethod
      */
     public const KLARNA_SLICE_IT = "klarnasliceit";
 
+    /**
+     * @link https://www.mollie.com/en/payments/mb-way
+     */
+    public const MB_WAY = "mb_way";
```
fjbender, line 128: The API identifier should be `mbway` so I would opt for the constant to be called `MBWAY` as well.

fjbender (changes requested):

src/Types/PaymentMethod.php
sandervanhooft: Yes, let's go with MBWAY here

sandervanhooft:
```suggestion
    public const MBWAY = "mbway";
```

Naoray (approved):

sandervanhooft:
Thanks @mweghorst !

sandervanhooft:
Release here: https://github.com/mollie/mollie-api-php/releases/tag/v2.79.0

PR #773, opened 2025-03-18
V3: Add psr/http-client

### LINKED ISSUE #772: Interface "Psr\Http\Client\ClientExceptionInterface" not found
When doing a call to the Mollie payment API (V3 beta) I get an error because the ClientExceptionInterface isn't found in vendor\mollie\mollie-api-php\src\Exceptions\RequestException.php.

Shouldn't psr/http-client be added in the require list of the mollie-api-php package?

sandervanhooft:
Fixes #772 

- Added `psr/http-client` dependency
- Bumped release version

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,6 +53,7 @@
     "ext-openssl": "*",
     "composer/ca-bundle": "^1.4",
     "nyholm/psr7": "^1.8",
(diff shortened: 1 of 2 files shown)


Naoray:
already done :-)

PR #780, opened 2025-04-10
Feat/766 filter out null payment methods with null status

### LINKED ISSUE #766: MethodEndpoint scope allActive is not accounting for methods with status:null
Today we've found out that there is a possibility Mollie marks a payment method as status:null instead of activated or deactivated.

We found this status on the method "voucher". Maybe when the business owner selected it before and then deactivated it that this status is set to null? (vouchers have mutliple submethods as far as i can see in the Mollie UI).

However when trying to sync the methods to the webshop we saw this voucher method coming back all the time while it's not activated in the Mollie UI.
When inspecting the API response we found that the voucher came back with the status property set to null.

`Mollie\Api\Resources\Method {#6868
        +resource: "method",
        +id: "voucher",
        +description: "Vouchers",
        +minimumAmount: {#6865
          +"value": "1.00",
          +"currency": "EUR",
        },
        +maximumAmount: {#6863
          +"value": "100000.00",
          +"currency": "EUR",
        },
        +image: {#6864
          +"size1x": "https://www.mollie.com/external/icons/payment-methods/voucher.png",
          +"size2x": "https://www.mollie.com/external/icons/payment-methods/voucher%402x.png",
          +"svg": "https://www.mollie.com/external/icons/payment-methods/voucher.svg",
        },
        +issuers: null,
        +pricing: null,
        +status: null,
        +_links: {#6861
          +"self": {#6894
            +"href": "https://api.mollie.com/v2/methods/voucher",
            +"type": "application/hal+json",
          },
        },
      },
`

We check the Mollie methods via the **allActive() method from the MethodEndpoint.php** before we make them available in the webshop. However this method does not seem to account for this null status and thus returns it as a valid payment method.

Maybe it's possible to improve this allActive() method to filter out the ones with a null status?

Naoray:
resolves #766

### DIFF
--- a/src/Http/PendingRequest.php
+++ b/src/Http/PendingRequest.php
@@ -51,8 +51,6 @@ class PendingRequest
         $this->method = $request->getMethod();
         $this->url = Url::join($connector->resolveBaseUrl(), $request->resolveResourcePath());
 
-        $this->middleware()->merge($request->middleware(), $connector->middleware());
-
         $this
             ->tap(new SetUserAgent)
             ->tap(new MergeRequestProperties)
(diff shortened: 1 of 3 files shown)


sandervanhooft:
Great! Can you also expose the filter flag here?

Naoray:
What do you mean? It s already exposed on the request, isn't it?

sandervanhooft:
Just discussed, let's leave it out there.

> Ah yes forgot to paste the actual line 🤦. I mean on the EndpointCollection class:
> 
> src/EndpointCollection/MethodEndpointCollection.php

sandervanhooft (approved):

Issue #784, opened 2025-04-16
ValidationException requires string now


RichardJong:
Since the upgrade to 3.0.0 we receive the following exception when there is a validation error:
```
Mollie\Api\Exceptions\ValidationException::__construct(): Argument #2 ($field) must be of type string, null given, called in /var/www/A/ftm/vendor/mollie  
  /mollie-api-php/src/Exceptions/ValidationException.php on line 34       
```

We get this error when creating some payments.

RichardJong:
Hi Sander, thanks for all the good work!

It seems that the API doesn't return a 'field' in a 422 response.

sandervanhooft:
Patch released. Thanks @RichardJong !

PR #785, opened 2025-04-16
Update ValidationException.php, make field optional

RichardJong:
`field` is not (always?) returned by the API, make this parameter optional.  See #784


Naoray:
Sorry. Saw yours after I already merged Sander's PR

PR #790, opened 2025-04-24
Allow psr/http-message:^1.0

jkabat:
As I'm using PAYUM on the top of the mollie library, I have problem to install stable 3.0 version with http-message:2.0 requirement.

Would be possible to allow older http-message which only misses return type hints?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,7 +55,7 @@
     "nyholm/psr7": "^1.8",
     "psr/http-client": "^1.0",
     "psr/http-factory": "^1.1",
-    "psr/http-message": "^2.0"
+    "psr/http-message": "^1.0|^2.0"
   },
   "require-dev": {
     "brianium/paratest": "^6.11",


sandervanhooft:
Hi @jkabat ,

Thanks for reaching out. Are you referring to https://github.com/PayHelper/payum-mollie ?

jkabat:
@sandervanhooft good catch, I didn't even know there is an implementation, but seems to be quite old and unmaintained.

I've meant the main library https://github.com/Payum/Payum for which I've created my own mollie gateway (its not public). I understand if you do not intend to support both version of http-message. In which case I will have to keep my fork up-to-date until the day Payum developers will get rid of old dependencies.

Naoray:
Hi @jkabat,

added it in v 3.0.4

jkabat:
@Naoray thank you!

PR #799, opened 2025-05-06
V2 fix/798 accepted response should accept empty body

### LINKED ISSUE #798: Release payment authorization endpoint has empty response body
### Bug Report

**Package:** mollie/mollie-api-php  
**Version:** v2.79.0  
**PHP Version:** 8.3
**Environment:** Symfony 6.4

---

#### Description

[The Release payment authorization endpoint](https://docs.mollie.com/reference/release-authorization) returns an empty response body for a test payment on local development with a `202` response code. This triggers a "No response body found." `ApiException`. 

---

#### Steps to Reproduce

```php
try {
    $this->api->payments->releaseAuthorization('some-payment-id');
} catch (Throwable $throwable) {
    // Handle errors
}
```

This eventually calls `\Mollie\Api\HttpAdapter\Guzzle6And7MollieHttpAdapter::parseResponseBody` which only allows a `204` response code with an empty response body causing the `ApiException` to be thrown. 

---

#### Suggested fix

Make sure the Release payment authorization endpoint returns `{}` in the response body [as described in the documentation](https://docs.mollie.com/reference/release-authorization).

Naoray:
closes #798

### DIFF
--- a/src/Endpoints/PaymentEndpoint.php
+++ b/src/Endpoints/PaymentEndpoint.php
@@ -187,15 +187,15 @@ class PaymentEndpoint extends CollectionEndpointAbstract
      *
      * @param Payment|string $paymentId
      *
-     * @return \stdClass
+     * @return void
      * @throws ApiException
      */
-    public function releaseAuthorization($paymentId)
+    public function releaseAuthorization($paymentId): void
     {
         $paymentId = $paymentId instanceof Payment ? $paymentId->id : $paymentId;
 
         $resource = "{$this->getResourcePath()}/" . urlencode($paymentId) . "/release-authorization";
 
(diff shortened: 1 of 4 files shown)


jvdhehvg:
@Naoray Thanks again for checking this issue. 
Your PR will solve the problem I mentioned, however it does not fix the fact that the endpoint does not return an empty json as stated in the documentation. If the API is fixed to reflect the documentation than this problem will be resolved without having to change the SDK.

Naoray:
@jvdhehvg the API returns an empty json which is translated as 'empty body' by our SDK. Therefore the `APIException` is thrown.

The issue is that a 204 response should never contain a response body, but a 202 can contain a response, but doesn't have to. The SDK assumed that if a response is empty it has to be accompanied by a 204 status, but that's not the case.

PR #802, opened 2025-05-17
Removes nullability from delete() method, as it cannot return null

Sjustein:
The cancel() method never returns null, thus delete can also not return null. It would be great if this fix could be merged, as it's causing typing issues in my codebase.


sandervanhooft:
LGTM. @Naoray ?

Naoray:
Yes... will release it at the start of next week with another fix.

PR #811, opened 2025-06-11
Fix: allow array of payment methods when creating a payment

jockri:
Based on the docs, the method parameter can also be an array, to limit the payment methods that are shown on the Mollie checkout page. 
[https://docs.mollie.com/reference/create-payment](https://docs.mollie.com/reference/create-payment)


jockri:
@Naoray I made the fix compatible with php 7.4

PR #812, opened 2025-06-16
Feat/add create webhook endpoint

Naoray:
adds webhook API endpoints as well as webhook event endpoint (s. https://docs.mollie.com/reference/create-webhook, https://docs.mollie.com/reference/get-webhook-event)


sandervanhooft (commented):
Some minor changes on the docs and I have added some recently introduced events, LTGM otherwise!

PR #820, opened 2025-07-24
Fix/819 signature date invalid

### LINKED ISSUE #819: Signature Date invalid format
I don't know if that's an API thing on your and or an SDK thing but:
When I try to create a payment mandate it always errors out with:
```
The signature date is invalid, please provide the signature date in Y-m-d format. The signature date cannot be before 1980
```
and then it gives me the payload it thinks it got:
```json
{"method":"directdebit","consumerName":"REDACTED","consumerAccount":"REDACTED","consumerEmail":"REDACTED","signatureDate":"2025-07-23T14:47:58+00:00","mandateReference":"REDACTED"}
```

This thing is, I know for a fact I'm passing `signatureDate: "2025-07-23"` (I've logged the payload, it's in there)

That didn't use to be an issue but it is now I guess and I'm not quite sure why...

Naoray:
fixes #819

The API has two different types of DateTime Formats.

- **Date** (Y-m-d)
	- create & update payment `dueDate`
	- `CreateCustomerPayment` -> `routing.releaseDate` & `dueDate`
	- `CreateMandateRequest` -> `signatureDate`
	- create & update subscription `startDate`
- **DateTime** (ISO 8601)
	- create payment link `expiresAt`

Previously we treated all the same and just formatted as DATE_ATOM in https://github.com/mollie/mollie-api-php/blob/v3.2.0/src/Utils/DataTransformer.php#L65-L67

With this PR we introduce our own wrapper classes around `DateTimeInterface` with `Mollie\Api\Http\Data\Date` and `Mollie\Api\Http\Data\DateTime` which have their format defined within the class.

For backwardscompatibility, all Requests listed in the list above accept both either the new `Date` or `DateTime` wrapper and `DateTimeInterface`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,25 @@ Starting with v3, all notable changes to this project will be documented in this
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased](https://github.com/mollie/mollie-api-php/compare/v3.2.0...HEAD)
+## [Unreleased](https://github.com/mollie/mollie-api-php/compare/v3.3.1...HEAD)
+
+## [v3.3.1](https://github.com/mollie/mollie-api-php/compare/v3.3.0...v3.3.1) - 2025-07-25
+
+### What's Changed
+
+* fix: signature validator handling null signatures by @Naoray in https://github.com/mollie/mollie-api-php/pull/822
+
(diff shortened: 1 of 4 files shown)


src/Http/Requests/CreateSubscriptionRequest.php
```diff
@@ -58,7 +62,7 @@ public function __construct(
         string $description,
         ?string $status = null,
         ?int $times = null,
-        ?DateTimeInterface $startDate = null,
+        $startDate = null,
```
sandervanhooft, line 65: Missing typehint, docBlock?

PR #829, opened 2025-09-19
Feat/add webhook mapping and events

Naoray:
- global Config that serves as a lookup map to easily map resources to their respective collection keys
- `MockEvent` to easily test event handling
- documentation on how to use `WebhookEventMapper` and `MockEvent` for testing webhooks
- `Str` utility class
- `classBasename` to `Utility`
- `WebhookEntity` to serve as Container for Resource data received through webhooks (-> can be transformed into BaseResource)
- Webhook Events that are instanced via the `WebhookEventMapper`

---

Currently waiting on a list of all Webhook Events


src/Webhooks/WebhookEventType.php
```diff
@@ -5,18 +5,69 @@
...
+            self::SALES_INVOICE_PAID,
+            self::PROFILE_CREATED,
+            self::PROFILE_VERIFIED,
+            self::PROFILE_BLOCKED,
+            self::PROFILE_DELETED,
+            self::ALL,
+        ];
+    }
+
+    public static function getAllNextGenWebhookEventTypes(): array
```
sandervanhooft: Why this method?

sandervanhooft: Yeah let's keep it clean.

sandervanhooft (approved):

PR #877, opened 2026-04-03
feat: add lines, addresses and minimumAmount to Payment Links API

Naoray:
## Summary

Mollie's Payment Links API now supports lines, billing/shipping addresses, and a minimumAmount — needed for Klarna and other BNPL methods. This wires those up in the SDK for both create and update.

- `CreatePaymentLinkRequest` and `UpdatePaymentLinkRequest` get four new optional params (`lines`, `billingAddress`, `shippingAddress`, `minimumAmount`), appended so existing callers aren't affected
- `PaymentLink` resource gets the matching properties
- Empty `lines: []` is coerced to null in both factories rather than sending an empty array to the API


sandervanhooft (approved):

.gitignore
```diff
@@ -8,3 +8,4 @@ php-scoper.phar
 php-scoper.phar.pubkey
 .php-cs-fixer.cache
 .phpunit.result.cache
+/.claude/skills/
```
sandervanhooft: this should go in local gitignore

PR #881, opened 2026-04-17
refactor(webhooks): bind connector to WebhookEventMapper for zero-arg asResource()

Naoray:
## Summary

Follow-up on #879. After Phase 2 landed, hydrating a webhook resource looks like this:

```php
\$event = (new WebhookEventMapper())->processPayload(\$body, \$signature);
\$payment = \$event->asResource(\$mollie);
```

The client is in scope when you set up the handler, so there's no good reason to pass it in again at the call site. Let the mapper remember it:

```php
\$mapper = new WebhookEventMapper([], \$mollie);
\$event  = \$mapper->processPayload(\$body, \$signature);
\$payment = \$event->asResource();   // no arg
```

Per-call override still works (`\$event->asResource(\$other)`), single-arg callers keep compiling, zero-arg on a mapper built without a connector throws a clear `LogicException` instead of a generic null dereference.

- `WebhookEventMapper::__construct()` takes an optional `?Connector` as its second argument and threads it onto every event it produces.
- `BaseEvent` stores the connector internally and uses it as the default for `asResource()`.
- `BaseEvent::asResource()` parameter becomes optional. Passing null with no bound connector throws `LogicException` with a message pointing at both fixes.

Tests cover all three paths: bound connector + zero-arg call works, explicit arg overrides the bound one, unbound mapper + zero-arg call throws.

## BC

Fully additive. The existing `asResource(Connector)` signature keeps compiling (the parameter is now optional, not removed). Zero runtime change for any caller that already passed the connector.

## Why it's separate

Kept out of #879 so that PR stays focused on the provenance refactor. This one is a quality-of-life tweak on top, easy to revert independently if it turns out to be the wrong ergonomics call.

Stacked on #879. Rebase onto `main` once that merges.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `WebhookEventMapper::processPayload()` gains an optional `?string
   $signature` parameter that is threaded through to the resulting event
   and carried onto hydrated resources via `WebhookSnapshotOrigin`.
+- `WebhookEventMapper::__construct()` accepts an optional `?Connector`
+  second argument. When supplied, events produced by the mapper carry
+  the connector so `BaseEvent::asResource()` can be called with zero
+  arguments (`new WebhookEventMapper([], $mollie)` → `$event->asResource()`).
+  An explicit argument to `asResource()` still overrides the bound one.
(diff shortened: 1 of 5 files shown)


Naoray:
Closing in favour of the capability-optional shape (withConnector() on the resource). Supersedes this approach.

PR #882, opened 2026-04-18
fix: change customer email property type to a nullable string

edwinvdpol:
Hi 👋 

After I added strict type to my PHP file, I got this error in production:

```
TypeError
strtolower(): Argument #1 ($string) must be of type string, null given
```
in this line of code:
```php
// $mollie is an instance of Mollie\Api\Resources\Customer
if ($user->email !== strtolower($mollie->email) && filled($user->email)) {
    ....
}
```
This PR changes the customer email property type to a nullable string.

### DIFF
--- a/src/Resources/Customer.php
+++ b/src/Resources/Customer.php
@@ -32,7 +32,7 @@ class Customer extends BaseResource
     public $name;
 
     /**
-     * @var string
+     * @var string|null
      */
     public $email;
 


sandervanhooft:
Thanks @edwinvdpol , let me verify this.

sandervanhooft (approved):

sandervanhooft:
Hi @edwinvdpol ,

Double check to ensure we get this right: this looks like an issue raised by a static type checker, i.e. PHPStan, correct? 

So the code itself is not actually throwing an exception?

edwinvdpol:
Hi @sandervanhooft,

It did throw an exception in production _after_ I added `declare(strict_types=1);` to the top of the file:

Screenshot 2026-04-24 at 10 28 07

PR #887, opened 2026-05-12
fix: add scopes query parameter to list customer mandates

### LINKED ISSUE #886: Add scopes query parameter support to Customer Mandates endpoint
That scopes option is missing in the mollie-api-php library:
`GET /v2/customers/{customerId}/mandates?scopes[]=customer-present`

Naoray:
## Summary

Closes #886.

`GET /v2/customers/{customerId}/mandates` accepts an optional `scopes[]` query parameter for filtering by usage context (e.g. `scopes[]=customer-present`) — see [list-mandates](https://docs.mollie.com/reference/list-mandates). The SDK didn't expose it, so callers had to drop down to the HTTP client to filter customer-present mandates.

- `GetPaginatedMandateRequest`: new optional `?array $scopes = null` constructor arg, merged into the request query.
- New `Mollie\Api\Types\MandateQuery` with `SCOPE_CUSTOMER_PRESENT` and `SCOPE_CUSTOMER_NOT_PRESENT` so callers don't have to know the literal strings.
- `MandateEndpointCollection::pageFor` / `pageForId` / `iteratorFor` / `iteratorForId` thread the parameter through. Added at the end of each signature to keep existing call sites untouched.

### DIFF
--- a/docs/recipes/mandates/manage-mandates.md
+++ b/docs/recipes/mandates/manage-mandates.md
@@ -29,12 +29,14 @@ try {
 
 ```php
 use Mollie\Api\Http\Requests\GetPaginatedMandateRequest;
+use Mollie\Api\Types\MandateQuery;
 
 try {
     // List all mandates for a customer
     $response = $mollie->send(
         new GetPaginatedMandateRequest(
-            customerId: 'cst_8wmqcHMN4U'
+            customerId: 'cst_8wmqcHMN4U',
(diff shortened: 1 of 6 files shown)


sandervanhooft (approved):

fjbender:
Should we maybe extend https://github.com/mollie/mollie-api-php/blob/main/docs/recipes/mandates/manage-mandates.md to reflect this change as well?

PR #888, opened 2026-05-12
feat: add Google Pay direct integration support

Naoray:
## Summary

Mollie added a Google Pay direct integration flow: collect the `googlePayPaymentToken` from the Google Pay JS SDK and send it on a regular create-payment call. The token is processed on Mollie's side via the credit card rails, so the request goes out with `method: creditcard` — not `googlepay`. The SDK already passes unknown payload keys through, so the token works today; what was missing was the `PaymentMethod::GOOGLEPAY` constant (so callers can surface Google Pay as a hosted-checkout option) and a test pinning the direct-integration round-trip.

- `src/Types/PaymentMethod.php`: adds `PaymentMethod::GOOGLEPAY = 'googlepay'`. Docblock notes the two flows — use this constant for hosted-checkout method selection; for direct integration set `method` to `PaymentMethod::CREDITCARD` and pass `googlePayPaymentToken`.
- `tests/Http/Requests/CreatePaymentRequestTest.php`: a direct-integration test that calls `$client->payments->create([...])` with `method: creditcard` + `googlePayPaymentToken` and asserts both keys land at the top level of the JSON body.
- `tests/Types/PaymentMethodTest.php`: pins the constant value at `'googlepay'`.

Ref: https://docs.mollie.com/docs/direct-integration-of-google-pay


tests/Types/PaymentMethodTest.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Tests\Types;
+
+use Mollie\Api\Types\PaymentMethod;
+use PHPUnit\Framework\TestCase;
+
+class PaymentMethodTest extends TestCase
```
sandervanhooft: Not sure how much value this assertion adds here.

sandervanhooft (approved):
One minor nit, lgtm otherwise!

PR #897, opened 2026-06-07
fix: correct @var type for Payment::status from string to PaymentStatus

### LINKED ISSUE #896: Type error for Payment::status
There's a type error in the docblock here:
https://github.com/mollie/mollie-api-php/blob/f3655ce92f88c01683f19d0454016749b94d3013/src/Resources/Payment.php#L99

Should be:
`@var \Mollie\Api\Types\PaymentStatus`
not string.

This causes errors in an IDE when trying to use this property.

Jah-yee:
Fixes #896 - changed @var string to @var \Mollie\Api\Types\PaymentStatus for the $status property (1 line fix).

### DIFF
--- a/src/Resources/Payment.php
+++ b/src/Resources/Payment.php
@@ -96,7 +96,7 @@ class Payment extends BaseResource implements EmbeddedResourcesContract
     /**
      * The status of the payment.
      *
-     * @var string
+     * @var \Mollie\Api\Types\PaymentStatus
      */
     public $status = PaymentStatus::OPEN;
 


seb303:
This PR is invalid and should be rejected.
See: https://github.com/mollie/mollie-api-php/issues/896

Jah-yee:
Thanks for the clarification @seb303 — I understand now. The @var string is correct since PaymentStatus constants are strings. My mistake for not verifying the actual type before submitting. Closing this PR. Thanks for your patience!

PR #901, opened 2026-06-10
v4.0.0: PHP 8.2+ modernization

### LINKED ISSUE #875: Improve static type inference for API client `send` method with strongly typed requests
In the v3 major release strongly typed request objects were introduced; see also https://github.com/mollie/mollie-api-php/blob/master/UPGRADING.md#typed-request-objects. Where these are a great addition for writing safer code, they unfortunately result in less strict inferred types when doing static analysis, as the return type of `send` is simply `mixed`:
https://github.com/mollie/mollie-api-php/blob/6feb9bbfede72eea26603e08f2ca987c5b3c2eec/src/Traits/SendsRequests.php#L40-L43

In https://github.com/mollie/mollie-api-php/issues/789#issuecomment-2824838380 I read:

> I'm planning to add more improvements on the returned resources over the coming versions.
> 
> If you have an idea on how we can improve the returned typing without adding a dedicated method on the request itself, please let me know!

I'm not sure if there are any ongoing efforts on this account, but wanted to create this issue mostly as an encouraging nod that such attempts would be greatly appreciated and as a way to track possible updates.

As for ideas, there's two roads forward I can see that come to my mind:
- leverage distinct methods, similar to how the 'old' interfaces worked, however this is seemingly (per https://github.com/mollie/mollie-api-php/issues/789#issuecomment-2824838380 again) not the intended way to go
- introduce a generic type `TResponse` to the `ResourceHydratableRequest`; in theory (I've not really attempted much of this), I think something along the lines of the following signature on `send` could then be used:
```php
    /**
     * @template TRequest of ResourceHydratableRequest
     * @template TResponse
     * 
     * @param TRequest
     * @return TRequest is ResourceHydratableRequest<TResponse> ? TResponse : mixed
     */
    public function send(Request $request)
```

### LINKED ISSUE #876: Add additional convenience creation methods for Money class
In the v3 major, various strongly types classed were introduced for typesafe handling of data, including the (very relevant) `Money` class. This removes various drawbacks from earlier versions. However, one is still required to provide very particularly formatted strings for money values, due to upstream API limitations. This is both easily susceptible to human error and moreover a likely layer that requires conversion in lots of applications, as one will rarely use strings to deal with amounts of money.

To ease this process, I would suggest adding one or more (static) methods to the `Money` class, that for example could do the following things:
- accept an integer amount (e.g. `Money::createFromCents`), then use say `number_format($amount / 100, 2, '.', '')` to correctly format this string
- accept a float amount (e.g. `Money::createFromFloat), then use say `number_format($amount, 2, '.', '')` to correctly format this string

For me personally, I would very much also like a similar method that would except a `Money` instance from [brick/money](https://github.com/brick/money) to create money here, say `Money::createFromBrickMoney($money) => new (amount: (string) $money->getAmount(), currency: $money->getCurrency()->getCurrencyCode())`, with this package then added as a suggestion to the composer depends. However, it is understandable if such a request is considered out of scope. 

Similarly, the [moneyphp/money](https://github.com/moneyphp/money) library also seems to be a widely used library for dealing with money inside PHP, which could justify adding a conversion method here for convenience.

Naoray:
This brings the v4 line into main. v4 modernizes the SDK for PHP 8.2+ with a documented path for integrations upgrading from v3. [UPGRADING.md](https://github.com/mollie/mollie-api-php/blob/e6ed44032c4b5450b1206772fef910c738136eab/UPGRADING.md) walks through every breaking change with before/after examples.

## What's new for users

**PHP 8.2+ required.** Support for 7.4–8.1 is dropped; we test against 8.2, 8.3 and 8.4.

**Native enums instead of type constants.** API value sets under `Mollie\Api\Types` are now string-backed enums (`PaymentStatus::Paid`). Status-style properties on resources are typed `EnumName|string`, so unknown values coming from the API degrade gracefully instead of crashing. The `GetAllConstants` trait is gone — enums bring `cases()` natively, and the three classes that relied on it keep a backwards-compatible `all()`.

**Fully typed resources and value objects.** Fields that used to be `\stdClass` are now real value objects — `$payment->amount` is a `Money`, and `$payment->amount->value` still works exactly as before. Value objects are `readonly`, constructors use property promotion (named arguments unchanged), and every file declares `strict_types=1`.

**Static analysers finally understand `send()`.** `MollieApiClient::send()` carries a generic return type, so PHPStan, Psalm and PhpStorm infer the concrete resource from the request class — manual `@var` casts can simply be deleted.

**A fluent Money builder.** `Money::of('EUR')->minorUnits(1099)` and `Money::of('EUR')->fromString('10.99')` handle currency-exponent quirks (JPY, BHD, …) for you, and `Money` is now `Macroable`, so custom factories no longer require subclassing.

**Smarter failure handling.** The new `ExponentialRetryStrategy` supports jitter and honours `Retry-After` on HTTP 429. `ValidationException` exposes per-field errors, `TooManyRequestsException` exposes the retry delay, and exception messages no longer leak API tokens or request bodies.

**Nicer test ergonomics.** `MockResponse` ships typed factories — `MockResponse::payment(...)`, `::customer(...)` and friends — so faked responses read like the API objects they imitate.

Profile webhook events (`ProfileCreated`, `ProfileVerified`, `ProfileBlocked`, `ProfileDeleted`) complete the event coverage, and the resource hydrator was rewritten on top of reflection to fill the new typed properties — origin handling (HTTP vs. webhook snapshot) behaves exactly as in v3.13.

## Parity with v3.14.0

The branch now contains main through v3.14.0. Payouts, terminal pairing codes, `applicationFee` on payment links, response origins, and webhook snapshots remain available. The zero-value request serialization fix is present, and `PaymentMethod::Wero` brings the latest v3 method addition to the v4 enum. The changelog includes the v3 release history and a prepared v4.0.0 entry.

## Compatibility

The breaking changes are deliberate and documented in [CHANGELOG.md](https://github.com/mollie/mollie-api-php/blob/e6ed44032c4b5450b1206772fef910c738136eab/CHANGELOG.md). The most likely touchpoints in real integrations are the enum case renames, `readonly` value objects blocking non-readonly subclasses (use `Macroable` instead), and `strict_types` surfacing implicit coercions. The updated branch passed 1,534 tests, PHPStan, formatting, and currency generation locally on PHP 8.4; CI covers PHP 8.2, 8.3 and 8.4.

Closes #875
Closes #876


src/Http/Data/MoneyBuilder.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Mollie\Api\Http\Data;
+
+readonly class MoneyBuilder
+{
+    public function __construct(
+        private string $currency,
+    ) {
+    }
+
+    public function minorUnits(int $amount): Money
```
sandervanhooft, line 14: Seems a bit inconsistent, `minorUnits()` vs `fromString()`.

`fromMinorUnits()`?

src/Http/Data/OrderLine.php
```diff
@@ -1,65 +1,28 @@
...
-        ?string $productUrl = null
+        public string $description,
+        public int $quantity,
+        public Money $unitPrice,
+        public Money $totalAmount,
+        public ?string $type = null,
+        public ?string $quantityUnit = null,
+        public ?Money $discountAmount = null,
+        public ?RecurringBillingCycle $recurring = null,
+        public ?string $vatRate = null,
```
sandervanhooft, line 20: Could make sense to introduce a VatRate object in order to enforce the correct formatting

PR #904, opened 2026-08-12
refactor(retry): name the Retry-After budget rule and harden RateLimit parsing

Naoray:
The Retry-After budget rule now lives in one millisecond-based helper used by both retry eligibility and delay calculation. Boundary coverage keeps those decisions aligned when maxDelayMs is not a whole second, while exponential and Retry-After calculations clamp before integer overflow can occur.

RateLimit parsing now handles comma-joined policy lists, selects matching policies, accepts zero-padded integers, and preserves valid sibling data when another member is malformed or contains an unknown extension. The retry documentation now states the maxDelayMs gate for HTTP 429 responses.


Naoray (commented):
src/Http/Data/RateLimit.php:147: `splitListMembers()` only checks the immediately preceding byte to decide whether a quote is escaped. A closing quote preceded by an even number of backslashes is unescaped, but this scanner treats it as escaped, stays inside the quoted string across the next top-level comma, and combines separate list members. For example, a state list whose first name ends in an escaped backslash followed by a valid `"second";r=2` member loses that second state, so it cannot match a valid `"second";q=3` policy. Please determine escaping from the parity of the consecutive backslash run before the quote and add a regression test proving the later member remains independently parseable.

PR #905, opened 2026-08-12
feat(retry): rate-limit aware retries and RateLimit header parsing

Naoray:
This backport targets the v3 maintenance branch and prepares **v3.15.0**. It adds response header accessors and parsed rate-limit metadata, exposes `Retry-After` delays through `TooManyRequestsException`, and introduces an opt-in exponential retry strategy with jitter and a configurable delay budget.

## Compatibility

The default `LinearRetryStrategy` and `RetryStrategyContract` remain unchanged, so existing custom strategies and subclasses keep working. Strategies that need exception-aware decisions, including HTTP 429 handling, can implement `ConditionalRetryStrategyContract`. The backport retains PHP 7.4 compatibility and leaves the inherited positional exception constructor arguments in place.

## Release preparation

The branch now includes v3 through v3.14.0, sets `CLIENT_VERSION` to `3.15.0`, and records these changes in the v3.15.0 changelog section.

## Verification

Local PHP 8.4: 1,205 tests and 3,445 assertions passed; PHPStan reported no errors. The updated PR CI runs on PHP 7.4 through 8.4.


src/Http/Data/RateLimit.php
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * @return array{name: string, parameters: array<string, int>}|null
+     */
+    private static function parseHeader(?string $header): ?array
+    {
+        if ($header === null) {
+            return null;
+        }
+
+        $segments = array_map('trim', explode(';', $header));
```
Naoray: Blocking: this is the pre-hardening parser rather than the behavior already merged on the v4 line. `Response::header()` uses PSR-7 `getHeaderLine()`, so multiple field values arrive comma-joined. This parser treats the complete value as one policy and returns `null` for valid inputs such as `"get-v2-refunds";r=3, "get-v2-payments";r=7` paired with `"post-v2-payments";q=5, "get-v2-payments";q=20`.

There is a second omitted hardening in `src/Http/ExponentialRetryStrategy.php:69`: the exponential value is cast to `int` before it is capped, so `delayBeforeAttemptMs(PHP_INT_MAX)` returns `0` instead of `maxDelayMs`.

Please port PHP 7.4-compatible versions of the merged v4 list splitting/policy matching and finite/capped exponential calculation, together with their regression tests.

sandervanhooft (commented):
Neat!

sandervanhooft (approved):

PR #924, opened 2026-09-07
Add wero as supported payment method

robindirksen1:
As mentioned https://docs.mollie.com/changelog/add-wero-as-a-supported-payment-method


sandervanhooft:
LGTM! Thanks @robindirksen1

PR #929, opened 2026-09-23
Run tests on PHP 8.5 and 8.6

jnoordsij:
PHP 8.5 has been stable for a while; it is also tested at the Laravel SDK (see https://github.com/mollie/laravel-mollie/commit/dd294ecd73622bb64b88d2f42d2a202c0c40c377).

PHP 8.6 has just received its first RC tag (see https://github.com/php/php-src/releases/tag/php-8.6.0RC2) and should therefore be stable enough for CI testing.

Also bump to 8.5 for other CI jobs.


jnoordsij:
Pending upstream support from paratest

